PR #384, opened 2025-01-27
fix: `XAccountController` should use streaming in `archiveBuild`

### LINKED ISSUE #383: Error importing large X archive
When importing a large X archive, it throws this error:

Image

This is Cyd 1.0.26.

It looks like the crash is happening in `JSON.stringify` here: https://github.com/lockdown-systems/cyd/blob/v1.0.26/src/account_x/x_account_controller.ts#L1244

After some quick searching, it looks like the problem is `JSON.stringify` can't handle this much data, or maybe the computer runs out of memory, and we need to update this to stream the JSON encoding instead of doing it all at once in memory.

redshiftzero:
Closes #383
I've tested the `archiveBuild` method using a small test archive

### DIFF
--- a/src/account_x/x_account_controller.ts
+++ b/src/account_x/x_account_controller.ts
@@ -1272,7 +1272,7 @@ export class XAccountController {
         await archiveZip.extract({ path: getAccountDataPath("X", this.account.username) });
     }
 
-    async writeJSONArray(streamWriter: fs.WriteStream, items: any[], propertyName: string) {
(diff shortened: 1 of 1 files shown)


micahflee (approved):
This looks like it should fix this specific problem to me. However, looking over this function (I'm the one who first wrote it) I can see how massive amounts of data combined with not much memory could cause it to crash in other ways too. I opened a future issue to deal with that: #385

PR #386, opened 2025-01-28
Fix small English grammar error in AccountXViewModel.ts

phrawzty:
It's a verb, in this case. :)


SaptakS (approved):
Thanks for your PR! Looks good to me.

PR #390, opened 2025-02-06
Better X loading page

micahflee:
When you first open Cyd and the first logged in account loads, there's a second of loading where you just a white screen with an hourglass in the message field.

This updates it so that it shows the Cyd Windows installer loading gif animation instead.

Screenshot 2025-02-05 at 5 54 42 PM


SaptakS (approved):
LGTM

PR #392, opened 2025-02-06
Make message dialogs look better

micahflee:
We open native dialogs with Electron's `dialog.showMessageBoxSync`. This supports passing in `message` but also `detail`, which is smaller text beneath the message. This PR adds support for using `detail`.

For example, when you're in open mode and you start Cyd, it used to pop this up:

Screenshot 2025-02-05 at 6 20 54 PM

Now it pops this up:

Screenshot 2025-02-05 at 6 19 50 PM


SaptakS (approved):
LGTM

Issue #380, opened 2025-01-26
Facebook support


clemlatz:
**Is your feature request related to a problem? Please describe.**

I wiped out three Twitter/X accounts using Cyd and it works great. Now I'd like to the the same with my Facebook account.

**Describe the solution you'd like**

Deleting all my Facebook data without deleting my account.

**Describe alternatives you've considered**

Doing it by hand.

**Additional context**

N/A

redshiftzero:
Proposed implementation steps: 

- [ ] Get Facebook accounts that are populated with data on the Facebook profile wall at minimum (and ideally Pages, Groups, comments) - WIP
- [ ] Add Facebook as a new platform to Cyd
- [ ] Save all posts (including media) on Facebook profile wall via the Cyd web view
- [ ] Delete posts on Facebook profile wall via navigating to each item in the Cyd web view and clicking "Move to Trash"
- [ ] Also support data import from Meta's "Download Your Information" export tool

micahflee:
I'm working on "Add Facebook as a new platform to Cyd" which includes defining the schema for Facebook settings. What settings should Facebook have? Considering none of the devs are active Facebook users, I feel like this is where some user research could be helpful...

For example, every post in Facebook has reactions including: like, love, care, haha, wow, sad, and angry.

![Image](https://github.com/user-attachments/assets/d501ea8d-8237-4b38-ab56-d19b9f839d87)

When deleting posts or comments, we likely want options like with X to delete based on timestamp, and based on reactions. Do people care about the type of reaction (e.g. keep posts with at least 20 loves) or not (e.g. keep posts with at least 20 reacts in total)?

For X, for tweets, the options are:

- Delete my tweets
  - older than N days
  - unless they have at least N retweets
  - or at least N likes

For Facebook, we could do:

- Delete my posts
  - older than N days
  - unless that have at least N reactions

Is this what most people would want?

micahflee:
FYI I'm starting to look at the HTTP requests that Facebook makes to figure out where the profile info is given to the client, and it looks like much of it is directly in the HTML inside tags like `<script type="application/json" data-content-len="155225" data-sjs>`.

redshiftzero:
> Cyd could just look for folders that start with ${accountID} and ignore the name component, and just use that.

This and option 2 seems like the best approach to me. Though perhaps it should be `~/Documents/Cyd/${accountID} - ${name}/` with a hyphen or another delimiter to make it clear there are two components

PR #395, opened 2025-02-08
Support for Facebook platform

micahflee:
Towards #380.

This adds the skeletons of Facebook support, with a FacebookAccountController on the main process with IPC functions hooked up, and a FacebookView and FacebookViewModel on the renderer side.

I've also implements login. It loads `https://www.facebook.com/` and then detects if the user is logged in or not by the presence of a `c_user` cookie set in the browser. If the user is logged in, it then collects all of the `<script type="application/json">` blocks from the HTML body, parses them all, and then searches them to find the user's Facebook account ID, name, and profile picture URI.

Login supports 2FA, including with U2F security keys, though there is a workaround because the "press your security key" thing doesn't pop up, like with X. If you login and 2FA is enabled, it does quickly flash this screen and then immediately move on:

Screenshot 2025-02-14 at 12 53 40 PM

Basically, Cyd detects you're logged in as soon as the `c_user` cookie is set, and so at this point the user is already logged in and it moves on from the login state. Do we maybe want to click the "Trust this device" button though? If so, we'd have to figure out how -- like, after the cookie is set, we need to wait for the page to finish loading and detect if we're at this page and if so, click the button. That can be left for a future PR though.

I also added very simple support for feature flags. To launch Cyd with Facebook enabled:

```sh
export CYD_FEATURE_FACEBOOK=1
npm run start
```

There's a new IPC function `isFeatureEnabled(feature: string): boolean` that just checks if `CYD_FEATURE_{feature.toUpperCase()}` is set to 1 or not. The way I'm using it right now is just in the `AccountView.vue` view -- it checks for Facebook and Bluesky features, and if the feature flags are set, it displays boxes for selecting those two types of accounts when adding a new account.

So if you run Cyd like:

```sh
export CYD_FEATURE_FACEBOOK=1
export CYD_FEATURE_BLUESKY=1
npm run start
```

When you add a new account it looks like this:

Screenshot 2025-02-14 at 1 01 48 PM

This all the feature flag does right now. So this means that everything else, including the migration to add the new `facebookAccount` table, and registering all the Facebook IPC functions, etc., is still enabled. If you've already added a Facebook account and then run Cyd again with `CYD_FEATURE_FACEBOOK=0`, you can use your Facebook account like normal without a problem.

Does this seem fine to ship in prod? I think so, but maybe I'm not thinking of something. My idea is once we have a feature (like Facebook) ready enough to enable by default, we just remove the code in `AccountView.vue` that only shows the Facebook option if the Facebook feature is enabled, so instead it always shows.

This PR adds the following:

- [x] Adds Facebook account to account selection page, along with FacebookView, FacebookViewModel, FacebookAccountController, and initial IPC functions
- [x] Supports logging into a Facebook account, and fetching the account ID, display name, and profile image
- [x] Supports Yubikeys, with a [custom U2F page for Facebook](https://cyd.social/docs-u2f-facebook/)
- [x] Supports Facebook error reports to the server
- [x] Includes a Facebook feature flag, so we can merge this earlier


redshiftzero (approved):
Ran through the Facebook and X (to check for regressions) flows and all looks good!

Issue #332, opened 2024-12-08
Support automatic scheduled runs


micahflee:
This was a vague plan from the beginning, but I got this feature request:

> I'm excited for the Bluesky features to come, but I wanted to register a vote for automation — that is, scheduled runs of Cyd on my machine — as the most useful feature you could add. I'm a journalist and I run Cyd to delete all tweets from four or more days ago, so I don't have an archive of old tweets that people can take out of context and weaponize to try to get me fired. My current routine is to run Cyd first thing every morning, but I have to remember to do that and sometimes I don't. Obviously a scheduled deletion would only work if my laptop is open, but it usually is. I would create an Automator or Shortcuts action to do it on my mac, but it appears Cyd needs to configure those actions itself which probably doesn't make sense for a cross-platform app.

I think this feature would require a few things:

- A system tray icon, probably which you can minimize/close Cyd to, so it can always be running in the background
- Some new UI flow for creating scheduled jobs, with options for indexing and deleting

micahflee:
Bluesky desktop parity sharpens the intended scheduling model and adds a notification requirement:

- Scheduling should be **reminder-driven**, not unattended deletion. A due schedule prompts the user to open the account, review the proposed save/delete work, and explicitly start it.
- Desktop reminders should use **local operating-system notifications**. They must not depend on Cyd server push notifications.
- Notification behavior should work with the desktop lifecycle we choose (for example, while Cyd is running/minimized); the UI must state clearly when Cyd cannot deliver reminders.
- Clicking a notification should deep-link to the relevant local account and its pending review flow.
- Multiple accounts and coincident reminders must remain distinguishable and must not create duplicate pending work.
- Imported archive settings must not silently create or enable schedules; schedules remain local client settings.
- Missing notification permission or OS support must degrade visibly without causing any account action to run automatically.

This intentionally differs from the original request for automatic scheduled runs: notification delivery is automation, but destructive account actions still require explicit user review and confirmation.

This requirement comes from the cross-platform Bluesky design documented in the repository glossary and ADR 0014.

_(This comment was written by an LLM)_

PR #403, opened 2025-02-17
if logged in, update user activity

redshiftzero:
This PR updates the Cyd server with usage data in compliance with Lockdown Systems' Privacy Policy. An API request gets sent after the Plausible analytics data to Cyd server:

When you run this PR against https://github.com/lockdown-systems/cyd-server/pull/128 you'll see:

```
api-1         | 02-17 20:45 werkzeug     INFO     172.19.0.1 - - [17/Feb/2025 20:45:30] "POST /user/activity HTTP/1.1" 200 -
```

### DIFF
--- a/src/cyd-api-client.ts
+++ b/src/cyd-api-client.ts
@@ -567,4 +567,22 @@ export default class CydAPIClient {
             return this.returnError("Failed to subscribe to newsletter. Maybe the server is down?")
         }
     }
+
+    // User activity
+
+    async postUserActivity(): Promise<boolean | APIErrorResponse> {
+        if (!await this.validateAPIToken()) {
+            return this.returnError("Failed to get a new API token.")
+        }
+
+        try {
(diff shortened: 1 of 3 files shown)


micahflee (changes requested):
This works great! Each time you open the Cyd app it updates the last accessed date.

However, if you open the Cyd app and you're not logged in, it should also update the last accessed date as soon as logging in is successful, somewhere around here: https://github.com/lockdown-systems/cyd/blob/main/src/renderer/src/modals/SignInModal.vue#L134-L154

Also, if we ever do #332, it's possible that someone might open Cyd and keep it open and actively in use for weeks but without updating the last accessed timestamp. Maybe there should be a `setInterval` that just automatically updates it (if logged in) every 6 hours or something. That could live in `App.vue`. So if the app is open in the background for days and days, we'll have the right day.

micahflee (approved):

PR #405, opened 2025-02-19
[facebook 1] feat: import wall posts from Meta export

### LINKED ISSUE #398: Facebook: support importing wall posts from an exported archive

### LINKED ISSUE #399: Facebook: build local Cyd archive website of Facebook data

### LINKED ISSUE #407: Facebook export: import re-shared posts
Followup to #405 
Related to #380

Import "Shared a post" (effectively like retweets in Facebook land) posts from the export file

### LINKED ISSUE #409: Facebook export: add media support
Followup to https://github.com/lockdown-systems/cyd/pull/405
Related to https://github.com/lockdown-systems/cyd/issues/380

Import media files referenced in the export file and display them on the generated static site

### LINKED ISSUE #410: Facebook export: add URL support
Followup to https://github.com/lockdown-systems/cyd/pull/405
Related to https://github.com/lockdown-systems/cyd/issues/380

Import URLs referenced in posts and display them on the generated static site

redshiftzero:
Closes #398
Closes #407 
Closes #409
Closes #399
Closes #410

TODO:
- [x] impl `importFacebookArchive`
- [x] Filter posts to import by "updated her status" in posts data file (without using text that will be in different languages)
- [x] File followup issues for other data types that can be imported from the Meta export file (one for "shared a post" in the posts file which is effectively like a retweet, one for "shared a group" which is sharing a group or group post), add media support, add URL support (currently commented out)
- [x] Only support JSON export

# Screenshots

# Add new import option
Screenshot 2025-02-20 at 8 09 20 PM

# Explain how to download archive
Screenshot 2025-02-19 at 3 42 10 PM

# Archive import
Screenshot 2025-02-19 at 6 34 16 PM

# Import finished
Screenshot 2025-02-20 at 10 29 41 AM

Now you can `select * from post` to view the posts


micahflee (changes requested):
There's a merge conflict with `main` right now that I haven't looked into, and also I noticed the packages changed enough that I had to `npm install` again, so might be good to merge `main` into this branch.

On `FacebookWizardImportPage.vue`, the instructions for downloading your archive could be improved/ I'm following them now. In my test account I actually don't need to prove my identity. Here's what I see when I download my Facebook archive:

Screenshot 2025-02-28 at 12 48 19 PM

I click "Download or transfer information" (you have a typo in the instruction, it should be "or" instead of "and").

Screenshot 2025-02-28 at 12 48 30 PM

There should be a step in the instructions for which to choose. I'm assuming "Available information".

Screenshot 2025-02-28 at 12 49 22 PM

There should be another step in the instructions to click "Download to device".

Screenshot 2025-02-28 at 12 50 09 PM

We should tell the user what to click her:
- Under "Date range" choose "All time"
- Notify should be an email you can check
- Format should be JSON (I assume? probably much easier to work with than HTML)
- Media quality: High

Then click "Create files".

Screenshot 2025-02-28 at 12 52 04 PM

The current instructions say "After requesting your archive, you'll need to wait at least a day, and maybe longer. Sorry!" -- that's true for X, but I'm not sure how long you have to wait for Facebook.

Anyway I'm currently waiting for my Facebook archive, but this is what I could test so far.

src/account_facebook/facebook_account_controller.ts
```diff
@@ -234,4 +253,264 @@ export class FacebookAccountController {
...
+                skipCount: skipCount,
+            };
+        }
+
+        // Import posts
+        if (dataType == "posts") {
+            const postsFilenames = await glob(
+                [
+                    // TODO: for really big Facebook archives, are there more files here?
+                    path.join(archivePath, "your_facebook_activity", "posts", "your_posts__check_ins__photos_and_videos_1.json"),
```
micahflee, line 408: Probably. We're going to need people with big Facebook accounts to help us test this. I actually discovered the same problem with twitter, that it splits JSON files into multiple files for people with like 100k tweets. We should ask around.

src/account_facebook/facebook_account_controller.ts
```diff
@@ -234,4 +253,264 @@ export class FacebookAccountController {
...
+                    for (const post of posts) {
+                        // Skip if no post text
+                        const postText = post.data?.find((d: { post?: string }) => 'post' in d && typeof d.post === 'string')?.post;
+                        if (!postText) {
+                            log.info("FacebookAccountController.importFacebookArchive: skipping post with no text");
+                            continue;
+                        }
+
+                        // Check if it's a shared post by looking for external_context in attachments
+                        const isSharedPost = post.attachments?.[0]?.data?.[0]?.external_context !== undefined;
```
micahflee, line 439: This is falsely classifying posts as shared posts. Here's one of the posts from my test account, which includes two links:

Screenshot 2025-03-03 at 5 32 09 PM

I added a commit to log the actual posts content when it skips a post (https://github.com/lockdown-systems/cyd/pull/405/commits/73d8821c4791c28c37b34a72441c5edc4a62c71d) and it skipped this as a shared post:

```
17:30:27.316 › FacebookAccountController.importFacebookArchive: skipping shared post {"timestamp":1740775067,"attachments":[{"data":[{"external_context":{"url":"https://stardewvalleywiki.com/JojaMart"}}]}],"data":[{"post":"One of the best wikis around: https://stardewvalleywiki.com/Stardew_Valley_Wiki\n\nAnd the store to boycott: https://stardewvalleywiki.com/JojaMart"},{"update_timestamp":1740775067},{},{},{}],"title":"Chase Westbrook shared a link."}
```

Here's the JSON:

```json
{
    "timestamp": 1740775067,
    "attachments": [
        {
            "data": [
                {
                    "external_context": {
                        "url": "https://stardewvalleywiki.com/JojaMart"
                    }
                }
            ]
        }
    ],
    "data": [
        {
            "post": "One of the best wikis around: https://stardewvalleywiki.com/Stardew_Valley_Wiki\n\nAnd the store to boycott: https://stardewvalleywiki.com/JojaMart"
        },
        {
            "update_timestamp": 1740775067
        },
        {},
        {},
        {}
    ],
    "title": "Chase Westbrook shared a link."
}
```

It appears that `external_context` is used for link previews, and probably other things too, even for posts that aren't reposts.

src/account_facebook/facebook_account_controller.ts
```diff
@@ -234,4 +253,264 @@ export class FacebookAccountController {
...
+                            log.info("FacebookAccountController.importFacebookArchive: skipping shared post");
+                            continue;
+                        }
+                        else if (post.attachments) {
+                            log.info("FacebookAccountController.importFacebookArchive: skipping unknown post type");
+                            continue;
+                        }
+
+                        postsData.push({
+                            id_str: post.timestamp.toString(),
```
micahflee, line 453: I think using the timestamp as an id is reasonable since (hopefully) this will be unique. But the bigger concern for me is, since this actually isn't the real ID of the post, how will we use the info we're storing to delete this post in the future, if we choose to?

For example, in my test account, if I click on a post it loads this URL:

`https://www.facebook.com/permalink.php?story_fbid=pfbid0UWD2yDkvpM5FoBCJFiG3wFvjKcCx45RaZFeTZm1EU7VzyyRXRNkKTjnYwWTKGqc6l&id=61572798227018`

The `id` part, `61572798227018`, is my Facebook ID. So the `pfbid0UWD2yDkvpM5FoBCJFiG3wFvjKcCx45RaZFeTZm1EU7VzyyRXRNkKTjnYwWTKGqc6l` part must be the post ID. (I stripped the `pfbid` from the beginning and tried base64 decoding it but nothing legible comes out.)

When I grep my FB archive for this string, there's nothing. So once we get to the point where we choose what posts to delete, how do we choose to delete this one if we can't build the URL to load it?

I don't think this is something we need to solve in this PR, but I do think it makes sense to consider what we actually get get from the archive. If we don't have real post IDs (or rather, story IDs, as it seems FB calls them) then maybe the deletion options just have to be like "delete all posts without media", "delete all videos", "delete all photos", etc., instead of basing it on reactions like we do with X.

micahflee (changes requested):
I think this is mostly good. It did skip importing some of the posts in my test archive because it thought they were shared posts when they're not though, and I also added a few comments on post IDs -- it's unclear to be me if it will be possible to query the local db for posts to delete if we don't have real IDs.

So, I could merge this now and you can fix the shared post thing in followup a PR, or you could fix it here. Up to you, but I'm comfortable merging this as is since it's behind a feature flag.

It might also be simpler, and require less migrations, to just try to tackle all of the links, media, shared posts, etc. all in the same PR. I bet these extra entities are somewhat related and might require overlapping work anyway.

src/account_facebook/facebook_account_controller.ts
SaptakS: Maybe we can check the domain to ensure that the `external_context.url` is a facebook link?

redshiftzero:
I'm going to merge all the Facebook PRs into this one to resolve conflicts and then address all issues here

redshiftzero:
> I see once the archive has been imported, we either have "Continue to Delete options" (which I guess is not implemented) or "Continue to Archive options". Continue to Archive options just takes us to a blank page, whereas I feel we can just redirect to the initial page for importing/building data?

Yep neither of these pages is implemented yet

src/account_facebook/facebook_account_controller.ts
redshiftzero: For my export, the reposts have the "data"->"external_context"->"url" but the URL is empty, so I've modified the logic to look for that

src/account_facebook/facebook_account_controller.ts
redshiftzero: I don't see how we can get the post ID from the archive so we may need to do the latter. We could also provide options to delete different types of content in a specified time period, e.g. "Delete all posts more than 2 years old"

redshiftzero:
Each entry in the JSON Array in the posts JSON file has a:
* `timestamp`
* `attachments` array
* `data` array
	* may contain an entry with `update_timestamp`
	* may contain an entry with `post` text
* `title`

Shares of groups have:
* The group name in the first entry of `attachments.data.0.name`
* The title will be "NAME shared a group"

Reposts have:
* The first entry of `attachments.data.0.external_context.url` is ""
* The title will be "NAME shared a post"
* There may _not_ be any text, but if there is, the first entry in the `data` array will be a the `post` text

A regular post:
* has `post` text in the `data` array
* The title will be "NAME updated her status"

An image post or video post:
* has a `media` entry in `attachments.data`
	* It includes exif data including upload IP
* has the `post` text in the `data` array
* The title will say "NAME added a new photo" or "NAME added a new video"

A life event:
* will have in the `attachments` array an entry with key `data` that has another array with the `life_event` key containing a date and title. Right now we're just pulling out the title and using it as the post text.

redshiftzero:
the link previews (stored in the export file) are now loaded into the database (#410) and rendered on the static site as part of this PR

micahflee (approved):
I've tested importing an archive and it works great. I realized that I need to actually wait for yet another archive because my current one doesn't include certain types of posts (including images).

What I've tested and works great:

- Post text
- Link previews

I'm happy to merge this as-is, but while I'm testing I just want to record everything I notice that should be handled in future PRs:

- The Facebook view sidebar should have a nav sort of like the X view sidebar does, so you can easily jump between the different views. (Of course, those different views don't exist yet. Just right now after importing an archive, there's no easy way to get back to the database view.)
- The FB static archive site does not embed media yet. (It successfully downloaded a video I posted to the `media` folder, but it's video isn't included in `assets/archive.json` and it's not embedded in the archive site.)
- Emote data (likes, etc.) is not included, but I guess maybe it's just not included in the archive?

What I haven't tested (because I need to wait for a new archive):

- import re-shared posts
- import posts with images, including multiple images

redshiftzero:
Regarding emote data, we don't have for the posts on the wall the number of likes/emotes etc. 
For the user's likes and reactions, we do have that content stored in: `your_facebook_activity/comments_and_reactions`

PR #411, opened 2025-02-20
[facebook 2] feat: add app to generate static site for facebook data

redshiftzero:
Stacked PR on #405 (i.e. when that is merged, this can be re-targeted to `main`)
Closes #399

~Still marked WIP because there's a test failure to debug~ seemed to be a flake, reran the test suite and all passed 🤔 

Screenshot 2025-02-20 at 10 24 33 PM

PR #415, opened 2025-02-21
[facebook 3] support reposts in Facebook archiving

### LINKED ISSUE #407: Facebook export: import re-shared posts

redshiftzero:
Stacked PR on #411
Closes #407 

This adds the repost data that exists in the JSON FB export to the database and displays it on the archive app:

Screenshot 2025-03-02 at 2 05 30 PM

### DIFF
--- a/src/account_facebook/facebook_account_controller.ts
+++ b/src/account_facebook/facebook_account_controller.ts
@@ -615,26 +615,3 @@ export class FacebookAccountController {
         };
     }
 }
-
-const getPostType = (element: Element): 'status' | 'shared_post' | 'shared_group' | 'other' => {
-    const pinDivs = element.querySelectorAll('._2pin');
-
-    if (pinDivs.length === 1) {
-        return 'status';
-    }
-
-    if (pinDivs.length === 2) {
-        // Check for group name structure
-        const firstPinContent = pinDivs[0].textContent?.trim();
-        if (firstPinContent && !firstPinContent.includes('div')) {
-            return 'shared_group';
-        }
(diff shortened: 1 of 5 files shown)


SaptakS:
Seems like this branch has merge conflicts

src/account_facebook/facebook_account_controller.ts
```diff
@@ -513,14 +533,11 @@ export class FacebookAccountController {
 
                         // Check if it's a shared post by looking for external_context in attachments
                         const isSharedPost = post.attachments?.[0]?.data?.[0]?.external_context !== undefined;
```
SaptakS, line 535: I think the logic for determining whether its a repost or if the post just includes other URLs need to be worked similar to the comment by @micahflee at https://github.com/lockdown-systems/cyd/pull/405#discussion_r1978472981

SaptakS:
I wonder if there is more data that can be stored regarding the repost, e.g. the reposted URL?

redshiftzero:
> I wonder if there is more data that can be stored regarding the repost, e.g. the reposted URL?
unfortunately in my export at least the URL field in external_context is not populated, so we don't have the URL of the reposted content, which is a bit sad

src/account_facebook/facebook_account_controller.ts
redshiftzero: Modified this per https://github.com/lockdown-systems/cyd/pull/405#discussion_r1979791144

PR #426, opened 2025-02-26
Refactor migrate to Bluesky code to fit into XViewModel

micahflee:
The "Migrate to Bluesky" code in the main branch works, but it's implemented differently than the rest of the X platform. This makes it so it works like the rest of the X platform.

Now when you start migrating, it runs the jobs `['migrateBluesky', 'archiveBuild']`  in `XViewModel`. And when you delete migrated posts, it runs the jobs `['migrateBlueskyDelete', archiveBuild'']`.

This also introduces the concept of `jobsType` which is the general type of jobs you're running. The valid options:

- `save`
- `archive`
- `delete`
- `migrateBluesky`
- `migrateBlueskyDelete`

Before, the various views would determine what type of job is currently running and therefore what text to display, what buttons to have, where to go next, etc., by looking in the XAccount settings options `saveMyData`, `archiveMyData`, and `deleteMyData`. This PR gets rid of all of that and instead replaces it with storing the `jobsType` in localStorage when jobs are supposed to start.

For example, if you're deleting things, on the Delete Options page you choose your settings and click Start Deleting. It will use the `setJobsType` function in `util.ts` to set it to `delete` for this account, which stores this info in localStorage. Then on the next page, which is XWizardReviewPage, it pulls the jobsType from localStorage to determine what info to display on this page -- all of the settings you just chose.

In this PR, I've updated it so when migrating tweets to Bluesky, and when deleting migrated tweets from Bluesky, it re-uses the progress and status components that the other jobs types use. I added new fields to the `XProgress` object to handle Bluesky migration jobs. And when running these jobs, it flashes the tweets on the screen that it's migrating (or deleting), just like it does when deleting tweets.

When a `migrateBluesky` or `migrateBlueskyDelete` finishes, the XFinishedRunningJobsPage page shows you what it did, shows you an error messages you received while doing it, and includes a link to view your Bluesky profile.

Misc to do list:

- [x] When displaying tweets (`XDisplayTweet`), it should replace t.co links with real links, and maybe display images/video, or at least express that they exist
- [x] When clicking cancel, it should bring you back to Migrate to Bluesky
- [x] Handle rate limit errors better (figure out how long to wait)
- [x] Handle emoji
- [x] Handle error: Record/text must not be longer than 300 graphemes

For the "Record/text must not be longer than 300 graphemes" issue, the problem is links. Links in tweets take 20 chars no matter the actual length of the tweet, but this is not true with atproto. For example I got this error migrating this tweet: `Snowden says "endpoint security is so terrifically weak" that NSA can often bypass crypto. He's talking about this. http://washingtonpost.com/world/national-security/us-spy-agencies-mounted-231-offensive-cyber-operations-in-2011-documents-show/2013/08/30/d090a6ae-119e-11e3-b4cb-fd7ce041d814_story.html` (302 chars). To handle it, I'm falling back to the original t.co links in this case, so it will be: `Snowden says "endpoint security is so terrifically weak" that NSA can often bypass crypto. He's talking about this. http://t.co/CdVFyxZKWF` (138 chars). It adds back X tracking, but at least it migrates the tweet.

This fixes most of the cases, but there are still others where the text is >300 chars anyway. For example if you import your tweets from an X archive and you have a quote tweet, it tacks the quote URL to the end of the tweet which could push it over the 300 character limit. For example, here's one of my tweets I tried to migrate: `Since Russia invaded Ukraine, @xxNB65 has been hacking the hell out of Russian agencies and companies and dumping all their data. They're using modified ransomware tools from Conti (pro-Putin Russian ransomware gang) to encrypt all their data but like, not selling back the keys https://t.co/Xkhp3wMjRt` (302 chars). It failed. I solve this case by shortening the text to less than 300 characters (splitting on words), and then making a new continuation post, like this:

Screenshot 2025-02-27 at 1 47 05 PM


src/renderer/src/views/x/XFinishedRunningJobsPage.vue
```diff
@@ -197,6 +211,74 @@ onMounted(async () => {
...
+                        <i class="fa-brands fa-bluesky bluesky-bullet" />
+                        <strong>{{ model.progress.migrateTweetsCount.toLocaleString() }}</strong> tweets to
+                        Bluesky
+                    </li>
+                </ul>
+
+                <div v-if="model.progress.migrateSkippedTweetsCount > 0" class="alert alert-warning mt-4">
+                    <p>
+                        <strong>
+                            {{ model.progress.migrateSkippedTweetsCount.toLocaleString() }} tweets
```
redshiftzero, line 228: nit: this will result in "1 tweets were skipped" (should have singular and plural cases, could punt to followup issue)

src/renderer/src/views/x/XDisplayTweet.vue
```diff
@@ -0,0 +1,123 @@
...
+                <div class="d-flex justify-content-center align-items-center">
+                    <div class="video-placeholder d-flex justify-content-center align-items-center">
+                        <span>Video Placeholder</span>
+                    </div>
+                </div>
+            </div>
+            <div v-if="model.runJobsState == RunJobsState.DeleteTweets"
+                class="d-flex mt-4 gap-3 justify-content-center align-items-center">
+                <div v-if="tweetItem.r">
+                    <i class="fa-solid fa-retweet" /> {{ tweetItem.r.toLocaleString() }} retweets
```
redshiftzero, line 78: same here: should have singular/plural cases to avoid "1 retweets"

redshiftzero (approved):
LGTM! works great, couple small nits inline

PR #430, opened 2025-02-28
Fixes gif downloading when building the archive

### LINKED ISSUE #421: Gifs are sometimes downloaded as jpg files (seemingly only the first frame)

**Describe the bug**
I'm trying to download my twitter "likes" along with their media files, and some of them contain gifs that are downloaded as jpg files.

For example, https://x.com/cheekybigfoot/status/1745694470660567074 

looks like this in my archive:

Image

with the following jpg files instead of a video or gif:

Image

(the file is named `16_1745694463274401792.jpg` and is ~8kB in the media folder)

**To Reproduce**
Choose build database from scratch, checking option "save my likes".

**Expected behavior**
It would be nice if all gif/videos were saved as such, and not to jpg sometimes.

 - OS: Windows 10
 - Version of Cyd: 1.0.28

SaptakS:
Fixes #421 

So seems like X saves the GIFs are video files when uploaded (or added from it's interface) as a part of the post. What actually gets rendered is a MP4 file instead of GIF, which is slightly sad since it would have been nice to save a GIF file for migrations elsewhere. However I have gone ahead and made changes to save the MP4 file created by twitter for the GIF.

This should help in storing the video information instead of just a JPG. I have also added tests and modified the local archive template code to render it correctly.

### DIFF
--- a/archive-static-sites/x-archive/src/components/TweetComponent.vue
+++ b/archive-static-sites/x-archive/src/components/TweetComponent.vue
@@ -61,7 +61,7 @@ function atUriToBlueskyUrl(atUri: string): string {
             <!-- Media -->
             <div v-if="tweet.media.length > 0">
                 <div v-for="media in tweet.media" v-bind:key="media.filename" class="mt-2">
-                    <template v-if="media.mediaType == 'video'">
+                    <template v-if="media.mediaType === 'video' || media.mediaType === 'animated_gif'">
                         <video controls class="img-fluid">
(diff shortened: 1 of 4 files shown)


micahflee:
I think the reason X converts GIFs to MP4 is so they can use the html <video> tag which is much more versatile than the  tag btw

micahflee (approved):
I tested this out and it works great!

I only tested with building an archive from scratch, not with importing one, but looking at the code I assume importing from archive should work the same. Did you test from an archive at all?

PR #434, opened 2025-02-28
Fix crash on signing into X related to getMediaPath

### LINKED ISSUE #433: Main branch can't add new X account
If you don't currently have an X account added and you add one, it crashes with this error:

```json
{
    "message": "The \"path\" argument must be of type string. Received null",
    "name": "TypeError",
    "stack": "TypeError [ERR_INVALID_ARG_TYPE]: The \"path\" argument must be of type string. Received null\n    at Object.join (node:path:1244:7)\n    at getAccountDataPath (/Users/user/code/cyd/.vite/build/main.js:2082:38)\n    at XAccountController.getMediaPath (/Users/user/code/cyd/.vite/build/main.js:85821:29)\n    at /Users/user/code/cyd/.vite/build/main.js:86421:31\n    at WebContents.<anonymous> (node:electron/js2c/browser_init:2:83740)\n    at WebContents.emit (node:events:519:28)"
}
```

This is because XView calls the `getMediaPath` IPC function, and it fails because the user hasn't logged into an account yet, so it can't build the media path.

This was just introduced in a different PR, so this bug hasn't been released yet.

micahflee:
Fixes #433

### DIFF
--- a/src/account_x/x_account_controller.ts
+++ b/src/account_x/x_account_controller.ts
@@ -3275,7 +3275,7 @@ export class XAccountController {
     }
 
     async getMediaPath(): Promise<string> {
-        if (!this.account) {
+        if (!this.account || !this.account.username) {
             return "";
         }
         const accountDataPath = getAccountDataPath("X", this.account.username);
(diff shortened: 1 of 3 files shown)


redshiftzero (approved):
Repro'd the initial issue described in #433 on main, and this resolves. LGTM!

PR #438, opened 2025-03-04
Handle rate limits while uploading media

micahflee:
In `blueskyMigrateTweetBuildRecord`, I've removed the try/catch blocks around `agent.uploadBlob` so that if it throws an error, the code in `blueskyMigrateTweet` can catch it and handle it like it does the other rate limit errors.

I've also made a `handleBlueskyAPIError` function to re-use, to avoid repeating myself each time we need to check for a Bluesky API error (uploading media in `blueskyMigrateTweetBuildRecord`, posting to Bluesky, posting the follow-up post to Bluesky, deleting a post from Bluesky).

Unfortunately, this PR is kind of hard to test. We'd have to trigger Bluesky rate limits, which means posting thousands of things to Bluesky. I did test it myself and successfully get a rate limit for posting (I still need to wait an hour for it to expire) but it's more difficult for me to trigger a `uploadBlob` rate limit. So I'm fine with not testing this ourselves, but instead rather making a dev release and have our tester test it.

### DIFF
--- a/src/account_x/x_account_controller.ts
+++ b/src/account_x/x_account_controller.ts
@@ -2900,7 +2900,7 @@ export class XAccountController {
 
             // Make sure it's not too big
             if (mediaData.length > maxSize) {
-                log.warn(`XAccountController.blueskyMigrateTweet: media file too large: ${tweetMedia[0].filename}`);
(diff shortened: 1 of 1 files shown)


redshiftzero (approved):
untested, but diff looks reasonable!

PR #442, opened 2025-03-05
Fix isArchiveOld logic

### LINKED ISSUE #441: "We recommend that you reimport your local database of tweets" displays when it shouldn't
Image

I built my database from scratch 3 months ago, but I imported it less than a minute ago. Since I recently imported it, it shouldn't show the warning.

micahflee:
Fixes #441 

The old logic returned true if either the last import or the last build was old. This now returns true only if one of them is old.


redshiftzero (approved):
makes sense!

PR #450, opened 2025-03-10
Allow bluesky handle input to start with @

### LINKED ISSUE #447: Inputting Bluesky handle doesn't support leading @-sign

**Describe the bug**

When you're working on setting up Bluesky, you have to put in your handle. Mine is `@free.law`. 

When I put that in, it didn't work. I thought that was because I use a custom domain, but it was because I should have put in `free.law` instead, without the at sign.

**To Reproduce**
1. Try to configure Bluesky
2. Click in the input box to add your handle.
3. Put in `@free.law` as your handle.

**Expected behavior**

It should work either way, with or without the @-sign.

 - OS: Linux
 - Version of Cyd: Don't know. See: https://github.com/lockdown-systems/cyd/issues/446

SaptakS:
Fixes #447 

This change just strips the @ from the beginning if the bluesky handle startswith @ and then sends the authorization request with the modified handle.

@micahflee I have not tested this with an actual bluesky account.

### DIFF
--- a/src/account_x/x_account_controller.ts
+++ b/src/account_x/x_account_controller.ts
@@ -2561,6 +2561,11 @@ export class XAccountController {
         }
 
         try {
+            // Check if the handle starts with @. If so, strip the @ and try authorizing
+            if (handle.startsWith("@")) {
+                handle = handle.substring(1);
+            }
+
             // Authorize the handle
             const url = await this.blueskyClient.authorize(handle);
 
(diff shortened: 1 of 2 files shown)


micahflee (approved):
I added a commit to do the same thing in the renderer (which technically means the IPC call to XAccountController shouldn't have a handle that starts with `@`, but it doesn't hurt to keep the check there). This is because when logging into a Bluesky account, it briefly displays the handle with two `@` symbols instead of one.

PR #461, opened 2025-03-13
sensitiveContextData should not be a bool

### LINKED ISSUE #455: Crash in runJobIndexMessages
We got this error report with this exception:

```json
{
  "exception": "TypeError: Cannot create property 'logs' on boolean 'true'"
}
```

Looking in the sensitive context data, it happened in `XViewModel.runJobIndexMessages`.

The error report doesn't include a line number or stack trace or anything (maybe we should improve this) but I'm pretty sure I've narrowed it down to this:

https://github.com/lockdown-systems/cyd/blob/main/src/renderer/src/view_models/XViewModel.ts#L1405-L1409

This called:

```js
await this.error(AutomationErrorType.x_runJob_indexMessages_Timeout, {
    exception: (error as Error).toString(),
}, true);
```

And passes in `true` where it should be passing in an object.

micahflee:
Fixes #455

### DIFF
--- a/src/renderer/src/view_models/XViewModel.ts
+++ b/src/renderer/src/view_models/XViewModel.ts
@@ -1405,7 +1405,7 @@ Please wait while I index all the messages from each conversation...`;
             if (!success) {
                 await this.error(AutomationErrorType.x_runJob_indexMessages_Timeout, {
                     exception: (error as Error).toString(),
-                }, true);
+                });
             }
 
             if (shouldSkip) {


src/renderer/src/view_models/XViewModel.ts
```diff
@@ -1405,7 +1405,7 @@ Please wait while I index all the messages from each conversation...`;
             if (!success) {
                 await this.error(AutomationErrorType.x_runJob_indexMessages_Timeout, {
                     exception: (error as Error).toString(),
-                }, true);
```
redshiftzero, line 1408: looking at the signature of the `BaseViewModel` `error` method:
`async error(automationErrorType: AutomationErrorType, errorReportData: any = null, sensitiveContextData: any = null, allowContinue: boolean = false)`, I'm wondering if the intent here was that this bool be the `allowContinue` field?

micahflee: Nope I don't think that was the intent. Because looking at the code, it loads the DM conversation URL and waits for messages to appear, and so if the message never appear, it shouldn't allow continue.

redshiftzero (approved):
KEWL

src/renderer/src/view_models/XViewModel.ts
redshiftzero: ahh ok, thanks

PR #462, opened 2025-03-13
Include full stack trace on errors

micahflee:
For our error reports, we've been collecting exception data just by running `toString()` on the Error, like:  `{ exception: (e as Error).toString() }`. But this ends up with error data like this:

```json
{
  "exception": "TypeError: Cannot create property 'logs' on boolean 'true'"
}
```

It would be much better if we could have actual stack traces, to pinpoint the file and line number where the error occurred. This PR adds a `formatError` function that returns a string that includes both the error message and the stack trace, and it updates all of the error report calls to use this instead.


redshiftzero (approved):
sique

PR #472, opened 2025-03-22
facebook: add `Delete` and `Archive` options

### LINKED ISSUE #465: facebook: implement "Continue to Archive Options" page
@SaptakS noted in [this comment](https://github.com/lockdown-systems/cyd/pull/405#issuecomment-2697621576) that "Continue to Archive Options" does nothing currently, because it's unimplemented. We should implement this similar to the X component.

redshiftzero:
Closes #465

This implements a few more of the required pages for Facebook after importing your Meta export:

Screenshot 2025-03-23 at 3 59 07 PM
Screenshot 2025-03-23 at 3 58 57 PM

### DIFF
--- a/src/account_facebook/facebook_account_controller.ts
+++ b/src/account_facebook/facebook_account_controller.ts
@@ -823,6 +823,7 @@ export class FacebookAccountController {
 
         // Count total posts
         const postsSaved: Sqlite3Count = exec(this.db, "SELECT COUNT(*) AS count FROM post", [], "get") as Sqlite3Count;
+        log.info('FacebookAccountController.getDatabaseStats: posts count:', postsSaved);
 
         // Count shared posts (reposts)
         const repostsSaved: Sqlite3Count = exec(this.db,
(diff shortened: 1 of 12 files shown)


micahflee (approved):
This looks good to me, though I think it's worth stepping back and figuring out exactly what we really need to implement for Facebook.

I'm actually starting to wonder if it makes sense to abandon importing form a Facebook export at all (in which case we'd also abandon the Archive step) if the export doesn't include what we need to do delete data (like post IDs)...

For X, there are three options for building the local database:

- Build database from scratch (Build)
- Import X archive (Import)
- Save HTML tweets, DMs, and/or bookmarks (Archive)

Archive and Build are exactly the same, except Archive has fewer options. If you do Build, there's no reason to do Archive. If you do Import, then Archive is there to get the rest of the stuff that wasn't included in the exported X archive.

Anyway, I guess it's hard to really know what we need to create until we learn how deleting works.

Issue #397, opened 2025-02-14
Facebook: support building database of wall posts from scratch


micahflee:

micahflee:
It's old, but here's some work that went into reverse engineering Facebook's GraphQL schema: https://github.com/CajuM/fb-graphql-schema

PR #482, opened 2025-03-28
Build FB database from scratch

micahflee:
Towards #397.

Changes to the renderer:

- I refactored `FacebookViewModel.loadFacebookURL` to rip out the complicated code about expected URLs and checking for redirects. That was added into X's view model because X was redirecting stuff, but so far I don't see a need for Facebook.
- I moved all of the code in `parseFacebookPostData` directly into the `runJobSavePosts` function, just because that's the only place it will be used.
- I updated the css selector for clicking the "Manage posts" button to not rely on the `aria-label` attribute, since that attribute is different depending on the user's locale.
- I updated the automation errors a bit. Before they had names like `facebook_runJob_clickManagePosts_URLChanged`, but to keep with the style of the other names, I changed it to `facebook_runJob_savePosts_URLChanged` -- we have a `savePosts` job but not a `clickManagePosts` job. I also added text strings to `AutomationErrorTypeToMessage` to give human-readable labels to these errors. And also I replaced the rate limit error with just a timeout error, since we don't actually know when a rate limit occurs yet.
- I made the IPC call `Facebook:saveGraphQLPostData` return a `FacebookProgress` object, and actually update it -- this way we can display progress in the renderer eventually, when we get to that part of the code.

Changes to the main process:

- In `FacebookAccountController`, I renamed `indexFacebookWallPostData` to `parseNode`. So far all it does is parse wall posts (or maybe more accurately, stories?), but I'd like this code to eventually parse every Node type we come across. (We should add support for parsing the User nodes too.)
- I renamed `indexFacebookWallPostMedia` to `parseAttachment` for similar reasons.
- In `parseGraphQLPostData`, I removed a check to only parse data when the response body includes the string `timeline_manage_feed_units`. This is because I can see how to fake this (like what if a user posts that string into a FB post). But also, I think it makes more sense to just parse _all_ nodes.

When I run this so far with my own test account it's crashing and I haven't finished figuring out why. My next task will be to start writing tests using my own GraphQL API responses, which I think will help me catch all of these bugs. But it's late on a Friday and so I figured I'd start a draft PR in case anyone wants to pick up where I left off.

To do:

- [ ] Start parsing User nodes too
- [ ] Add tests with my own GraphQL response data


SaptakS (approved):
Looks good to me. The only thing I am slightly thinking which I feel can be addressed in separate PR is if we want to try to save user for every node parsing, because I am guessing for a user, most of the time it will be same user for most stories. But we are doing like 2 DB queries per node.

SaptakS:
I feel we should just merge this PR so we have the basic structure and can work parallely on the remaining issues that are needed for making the MVP release.

PR #485, opened 2025-04-01
If loading a URL fails, wait 1 second before retrying loading the URL

### LINKED ISSUE #466: Error invoking remote method 'GUEST_VIEW_MANAGER_CALL'
We've gotten multiple error reports that include the error `Error invoking remote method 'GUEST_VIEW_MANAGER_CALL'`, including:

```json
{
  "url": "https://x.com/login",
  "exception": "Error: Error invoking remote method 'GUEST_VIEW_MANAGER_CALL': Error: ERR_TIMED_OUT (-7) loading 'https://x.com/login'"
}
```

```json
{
  "exception": "Error: Error invoking remote method 'GUEST_VIEW_MANAGER_CALL': Error: ERR_ABORTED (-3) loading 'about:blank'",
  "state": "WizardDeleteOptions",
---snip---
}
```

```json
{
  "exception": "Error: Error invoking remote method 'GUEST_VIEW_MANAGER_CALL': Error: ERR_ABORTED (-3) loading 'about:blank'",
  "state": "WizardImportStart",
  "jobs": [],
  "currentJobIndex": 0
}
```

```json
{
  "exception": "Error: Error invoking remote method 'GUEST_VIEW_MANAGER_CALL': Error: Script failed to execute, this normally means an error was thrown. Check the renderer console for the error."
}
```

This appears to be intermittent, and I have noticed it myself in the past, though I don't know how to reliably reproduce it.

Some potentially useful links for further reference:

- https://github.com/electron/electron/issues/24171
- https://www.electronjs.org/docs/latest/breaking-changes#default-changed-enableremotemodule-defaults-to-false
- https://nornagon.medium.com/electrons-remote-module-considered-harmful-70d69500f31

micahflee:
Hopefully fixes #466, but it's hard to tell because it's hard to reliably reproduce.

I believe the GUEST_VIEW_MANAGER_CALL error happens because we're trying to load a URL in the `<webview>` before something is fully initialized (either the webview itself, the MITM CA, or something).

Our `BaseViewModel.loadURL` function tries 3 times before erroring it, but it doesn't wait between retries. This sleeps 1 second between retries, which hopefully will give the webview or MITM proxy, or whatever, enough time to finish initializing.

See https://github.com/lockdown-systems/cyd/issues/466#issuecomment-2770640629 for more details.

### DIFF
--- a/src/renderer/src/view_models/BaseViewModel.ts
+++ b/src/renderer/src/view_models/BaseViewModel.ts
@@ -483,6 +483,9 @@ export class BaseViewModel {
                                 this.sleep(1000);
                             }
                         }
+                    } else {
+                        // Wait 1 second before retrying
+                        this.sleep(1000);
                     }
                 }
             }


SaptakS (approved):
I mean I can't reproduce, but seems logical to try.

PR #489, opened 2025-04-02
DRY up the release scripts

micahflee:
There are several different permutations of how we make a Cyd build: platform (windows, mac, linux), mode (local, prod, dev), command (make or publish), and before there was a different script for each of these. The Mac and Linux scripts were in bash and the Windows scripts were in PowerShell.

This merges them all together into a single Node script, `make.js`, that takes args uses logic to avoid all the duplication.


redshiftzero (approved):
utACK

PR #493, opened 2025-04-08
Make X SQL queries considerably more efficient

### LINKED ISSUE #428: Linux App Locks System Up with Huge Amounts of vRam Usage

**Describe the bug**
When the Linux app is used, it causes my system to slow down massively even when an archive is not yet being imported.

When I was importing a large archive, the rest of my system was unusable.

**To Reproduce**
1. Load the Linux App in Mint
2. Load System Monitor and look at the numbers.
3. My personal third step was to try and load a computer game, which simply did not manage to complete.

(Double-check Performed: Once I closed Cyd and all the processes went away, I could use my system as normal)

**Expected behavior**
Being able to use the app while doing other things.

**Screenshots**
[Amounts of VRAM Example](https://cdn.bsky.app/img/feed_fullsize/plain/did:plc:pyyjqvl3ihwsry3zqdb4cqf4/bafkreic4n62stdobdwqrcvwh3dakwz6s47dyfyk3yc6fhxhvmah2nmmavi@jpeg)

 - OS: Linux Mint 21.3
 - Version of Cyd: 1.0.28 (I downloaded it today)

**Additional context**
Not yet, but if you need me to find error logs, please let me know which ones are the most useful for this.

### LINKED ISSUE #492: UNIQUE constraint failed: tweet_media.mediaID
[This comment](https://github.com/lockdown-systems/cyd/issues/428#issuecomment-2784877380) hints at an unrelated bug due to a unique constraint violating in the SQL schema. Here's the error log:

```
(node:48369) UnhandledPromiseRejectionWarning: Error: {"exception":{"message":"UNIQUE constraint failed: tweet_media.mediaID","name":"SqliteError","stack":"SqliteError: UNIQUE constraint failed: tweet_media.mediaID\n    at u (/usr/lib/cyd/resources/app.asar/.vite/build/main.js:11:623)\n    at /usr/lib/cyd/resources/app.asar/.vite/build/main.js:336:9153"},"sql":"INSERT INTO tweet_media (mediaID, mediaType, url, filename, startIndex, endIndex, tweetID) VALUES (?, ?, ?, ?, ?, ?, ?)","params":["1048277200338915328","animated_gif","https://t.co/lJSM4BtQZK","1048277200338915328.mp4","54","77","1048277315183169540"]}
   at u (/usr/lib/cyd/resources/app.asar/.vite/build/main.js:11:670)
   at /usr/lib/cyd/resources/app.asar/.vite/build/main.js:336:9153
(Use cyd --trace-warnings ... to show where the warning was created)
```

Here's the exception, easier to read:

```json
{
    "exception": {
        "message": "UNIQUE constraint failed: tweet_media.mediaID",
        "name": "SqliteError",
        "stack": "SqliteError: UNIQUE constraint failed: tweet_media.mediaID\n    at u (/usr/lib/cyd/resources/app.asar/.vite/build/main.js:11:623)\n    at /usr/lib/cyd/resources/app.asar/.vite/build/main.js:336:9153"
    }
}
```

Here's the stack, easier to read:

```
SqliteError: UNIQUE constraint failed: tweet_media.mediaID
    at u (/usr/lib/cyd/resources/app.asar/.vite/build/main.js:11:623)
    at /usr/lib/cyd/resources/app.asar/.vite/build/main.js:336:9153
```

Here's the SQL query:

```
INSERT INTO tweet_media (mediaID, mediaType, url, filename, startIndex, endIndex, tweetID) VALUES (?, ?, ?, ?, ?, ?, ?)
```

And the params:

```
["1048277200338915328","animated_gif","https://t.co/lJSM4BtQZK","1048277200338915328.mp4","54","77","1048277315183169540"]
```

micahflee:
Hopefully this will fix #428 and #492.

This PR should make the performance of using the X platform waaaaaay better.

## Only run `getDeleteReviewStats` IPC function when it needs to

Before, when an X account tab was loaded, it created an `XViewModel` object, which runs `refreshDatabaseStats` to update its internal state of the database stats. This included a call to the `getDeleteReviewStats` IPC function, which is used for the review page when you're getting ready to delete. It basically does some heavy database queries to learn the number of tweets, retweets, likes, etc. it's going to delete, based on your settings.

But `getDeleteReviewStats` only needs to get run when you get to the review page, not when a `XViewModel` is initialized, and not every time `refreshDatabaseStats` is called either.

This PR removes `getDeleteReviewStats` from `refreshDatabaseStats`, and instead manually calls it in the `onMounted` function of `XWizardReviewPage.vue`. So basically, the very resource-intensive IPC function is only called right when you're getting ready to review your changes, and not when you first open the app.

## Use 1 query instead of 1+2N queries, so it's a bazillion times faster if you have lots of data

The `getDeleteReviewStats` IPC function learns how many tweets, retweets, likes, etc. it's going to delete by calling the following functions in `XAccountController`:

- `deleteTweetsStart`
- `deleteRetweetsStart`
- `deleteLikesStart`
- `deleteBookmarksStart`

Each of these returns a `XDeleteTweetsStartResponse` object, which is this:

```ts
export type XDeleteTweetsStartResponse = {
    tweets: XTweetItem[];
}
```

Where `XTweetItem` is an object that includes details about the tweet, and also array of filenames for images and video embedded in the tweet:

```ts
export type XTweetItem = {
    id: string; // tweetID
    t: string; // text
    l: number; // likeCount
    r: number; // retweetCount
    d: string; // createdAt
    i: string[]; // image filenames
    v: string[]; // video filenames
}
```

These functions used to be implemented really badly. They would basically do this:

- Select all the rows from `tweet` that need to be deleted in (1 query)
- Loop through those rows, and for each row, select all the media from that tweet (another query), and all URLs from that tweet (yet another query)

So if you have 1000 tweets, that would be 2001 query (1 query to select the tweets, 1000 queries to select the media, and another 1000 queries to select the URLs). And if you have 100k tweets, that's 200k SQL queries.

I've refactored it so that it select the tweets, media, and URLs in just one SQL query. So if you have 100k tweets, it now makes 1 SQL query instead of 200k, lol.

## Use `INSERT OR REPLACE INTO` queries instead of separate `DELETE FROM` and `INSERT INTO` queries

I also believe this PR fixes #492, though I'm not entirely sure why that was happening to begin with.

But basically, all over the place, before inserting a row into the database, I'd make a query to see if that row already exists, and if it does, I'd do another query to delete it. Then, I'd do a query to insert it. All of this can be done with a single `INSERT OR REPLACE INTO` query.

So basically I updated it to change code like this:

```ts
// Have we seen this media before?
const existing: XTweetMediaRow[] = exec(this.db, 'SELECT * FROM tweet_media WHERE mediaID = ?', [media["media_key"]], "all") as XTweetMediaRow[];
if (existing.length > 0) {
    // Delete it, so we can re-add it
    exec(this.db, 'DELETE FROM tweet_media WHERE mediaID = ?', [media["media_key"]]);
}

// Index media information in tweet_media table
exec(this.db, 'INSERT INTO tweet_media (mediaID, mediaType, url, filename, startIndex, endIndex, tweetID) VALUES (?, ?, ?, ?, ?, ?, ?)', [
    media["media_key"],
    media["type"],
    media["url"],
    filename,
    media["indices"]?.[0],
    media["indices"]?.[1],
    tweetLegacy["id_str"],
]);
```

Into code like this:

```ts
// Index media information in tweet_media table
exec(this.db, 'INSERT OR REPLACE INTO tweet_media (mediaID, mediaType, url, filename, startIndex, endIndex, tweetID) VALUES (?, ?, ?, ?, ?, ?, ?)', [
    media["media_key"],
    media["type"],
    media["url"],
    filename,
    media["indices"]?.[0],
    media["indices"]?.[1],
    tweetLegacy["id_str"],
]);
```

This should fix #492 because `INSERT OR REPLACE INTO` shouldn't fail on unique constraint errors -- it should just replace the row.

### DIFF
--- a/src/account_x/x_account_controller.ts
+++ b/src/account_x/x_account_controller.ts
@@ -444,6 +444,65 @@ export class XAccountController {
         log.info("XAccountController.initDB: database initialized");
     }
 
+    // Helper function to fetch tweets with media and URLs
+    private fetchTweetsWithMediaAndURLs(
+        whereClause: string,
+        params: (string | number)[]
+    ): XTweetItem[] {
+        const query = `
+            SELECT
+                t.tweetID, t.text, t.likeCount, t.retweetCount, t.createdAt,
+                tm.mediaType, tm.filename AS mediaFilename,
+                tu.expandedURL AS urlExpanded
+            FROM tweet t
+            LEFT JOIN tweet_media tm ON t.tweetID = tm.tweetID
+            LEFT JOIN tweet_url tu ON t.tweetID = tu.tweetID
+            WHERE ${whereClause}
+            ORDER BY t.createdAt ASC
+        `;
+
+        const rows = exec(this.db, query, params, "all") as {
+            tweetID: string;
+            text: string;
+            likeCount: number;
+            retweetCount: number;
+            createdAt: string;
+            mediaType: string | null;
+            mediaFilename: string | null;
+            urlExpanded: string | null;
+        }[];
+
+        // Group the results by tweetID
+        const tweetMap: Record<string, XTweetItem> = {};
+        for (const row of rows) {
+            if (!tweetMap[row.tweetID]) {
+                tweetMap[row.tweetID] = {
+                    id: row.tweetID,
+                    t: row.text.replace(/(?:\r\n|\r|\n)/g, '<br>').trim(),
+                    l: row.likeCount,
+                    r: row.retweetCount,
+                    d: row.createdAt,
+                    i: [],
+                    v: [],
+                };
+            }
+
+            // Add media files
+            if (row.mediaType === "photo") {
+                tweetMap[row.tweetID].i.push(row.mediaFilename!);
+            } else if (row.mediaType === "video") {
+                tweetMap[row.tweetID].v.push(row.mediaFilename!);
+            }
+
+            // Replace URLs in the text
+            if (row.urlExpanded) {
+                tweetMap[row.tweetID].t = tweetMap[row.tweetID].t.replace(row.urlExpanded, row.urlExpanded);
+            }
+        }
+
+        return Object.values(tweetMap);
+    }
+
     resetProgress(): XProgress {
         log.debug("XAccountController.resetProgress");
         this.progress = emptyXProgress();
@@ -1612,6 +1671,8 @@ export class XAccountController {
 
     // When you start deleting tweets, return a list of tweets to delete
     async deleteTweetsStart(): Promise<XDeleteTweetsStartResponse> {
+        log.info("XAccountController.deleteTweetsStart");
+
         if (!this.db) {
             this.initDB();
         }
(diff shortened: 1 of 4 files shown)


src/account_x/x_account_controller.ts
```diff
@@ -444,6 +444,65 @@ export class XAccountController {
...
+            SELECT
+                t.tweetID, t.text, t.likeCount, t.retweetCount, t.createdAt,
+                tm.mediaType, tm.filename AS mediaFilename,
+                tu.expandedURL AS urlExpanded
+            FROM tweet t
+            LEFT JOIN tweet_media tm ON t.tweetID = tm.tweetID
+            LEFT JOIN tweet_url tu ON t.tweetID = tu.tweetID
+            WHERE ${whereClause}
+            ORDER BY t.createdAt ASC
+        `;
```
SaptakS, line 462: I think we should ensure we use more joins for such situations throughout our codebase instead of needing to do multiple queries. 💯

src/account_x/x_account_controller.ts
```diff
@@ -625,7 +627,7 @@ export class XAccountController {
         }
 
         // Add the tweet
-        exec(this.db, 'INSERT INTO tweet (username, tweetID, conversationID, createdAt, likeCount, quoteCount, replyCount, retweetCount, isLiked, isRetweeted, isBookmarked, text, path, hasMedia, isReply, replyTweetID, replyUserID, isQuote, quotedTweet, addedToDatabaseAt) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)', [
+        exec(this.db, 'INSERT OR REPLACE INTO tweet (username, tweetID, conversationID, createdAt, likeCount, quoteCount, replyCount, retweetCount, isLiked, isRetweeted, isBookmarked, text, path, hasMedia, isReply, replyTweetID, replyUserID, isQuote, quotedTweet, addedToDatabaseAt) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)', [
```
SaptakS, line 630: I think in sqlite you can just write `REPLACE` instead of `INSERT OR REPLACE` and it basically does the same thing. Though I guess this is slightly more readable.

PR #501, opened 2025-04-15
Let users know about Premium, Cyd for Teams, and donations

### LINKED ISSUE #497: Upsell after finishing jobs
We should add a Vue component that's basically an ad for Premium, and maybe also Cyd for Teams. If the user is logged in and already has premium access, it should hide it. Otherwise, it should display it.

We should add this component to XFinishedRunningJobsPage, and, when it's ready, the Facebook equivalent.

micahflee:
Fixes #497.

This changes slightly depends on https://github.com/lockdown-systems/cyd-server/pull/141.

When reviewing, it would be helpful to confirm that the actual state changes are correct, like when going from not signed in to signed in, when going from no premium to premium, etc. You can test on the dev server (`npm run config-dev`) or spin up a local server to test (`npm run config-local`). If you use a local server, you'll need to use stripe-cli to forward stripe webhooks.

This PR creates a new component called `UpsellComponent.vue` and includes it on `XFinishedRunningJobsPage.vue`. When you run any jobs in an X account (like saving tweets), this is the page that it shows when the jobs are finished. Here's how it works, with all the different states:

I've finished running jobs, and I'm not signed into a Cyd account at all:

Screenshot 2025-04-14 at 5 00 29 PM

When you mouse-over the Premium, Cyd for Teams, or Donate boxes, the cursor becomes a pointer and they are clearly highlighted.

Clicking Donate opens https://opencollective.com/lockdown-systems in your default browser.

At this state, clicking Premium or Cyd for Teams opens the sign in modal:

Screenshot 2025-04-14 at 5 02 15 PM

When you successfully sign in, it automatically opens the Cyd dashboard in your default browser, which looks like this:

Screenshot 2025-04-14 at 5 04 49 PM

Back in the Cyd window, the text slightly changes. Under Premium it now says "Upgrade to Premium" and under Cyd for Teams it now says "Start a Team". So this is basically the state where the user is signed in to a Cyd account but does not have an active premium plan:

Screenshot 2025-04-14 at 5 05 41 PM

Now, if you click Premium it loads the Cyd dashboard, and if you click Cyd for Teams, it goes into the Cyd for Teams page to create a new team, like this:

Screenshot 2025-04-14 at 5 06 29 PM

That's what the related PR (https://github.com/lockdown-systems/cyd-server/pull/141) does, it makes it so you can click a link and be signed in and start at the teams page.

Now, if you go back to the Cyd dashboard and pay for a Premium plan, it will look like this:

Screenshot 2025-04-14 at 5 07 55 PM

Back in Cyd, the upsell component doesn't refresh yet. But in the dashboard if you click "Open Cyd to Get Started", and then open the cyd:// link, this now forces a refresh of state.

So, here is the state where the user has is signed in and has an individual Premium plan:

Screenshot 2025-04-14 at 5 09 31 PM

The text here thanks the user, and also lets them know about Cyd for Teams (links to docs) and accepting donations (links to OpenCollective).

Here's the state for users who are signed in and are already part of a Cyd for Teams plan, not an individual plan:

Screenshot 2025-04-14 at 5 11 58 PM

In this case, it only tells the user about donations (links to OpenCollective).

When reviewing, please give feedback on:

- the language used
- the design
- and of course the code

Question: should we display prices here?


SaptakS:
This is good, but I feel like the design is aggressive and very close to [deceptive design patterns](https://www.deceptive.design/). To me, from the design, it feels like the only next steps available to the user is to pay or donate and is not clear what are the actions they can still take without paying. I feel it should feel like an additional option but also like we are okay if thy only wanted the free features.

I feel often, it's a dismissable component, and not the main attracting component. Maybe we can also somehow separate the component of "We are open source and would love your support so please donate" from "Want more features? Try Cyd Premium, or get your entire team to use Cyd with Cyd for teams"... Of course someone needs to write better copy than this.

micahflee:
I've made some changes to make it way more subtle and optional-looking.

First, the upsell component is now below the buttons (so the Continue button is obvious).

I also changed the colors so it's all white background and greys, instead of primary action colors. Here's what it looks like now:

Screenshot 2025-04-15 at 9 50 07 AM

And here's what it looks like if you're already on Premium:

Screenshot 2025-04-15 at 9 56 31 AM

I also fixed a bug I discovered related to signing out of one account and signing into another. When you sign into a new account, it didn't refresh the API token, so all API requests were still using the old token from the first account.

SaptakS:
I like the change, in the sense it gets away from the deceptive design. But I now feel it's too subtle and not much differentiating from the stats containers. I would maybe change the border color of the container and the Premium Button to the same colour as the "Browse Archive" button, to make it stand out a little more.

micahflee:
I made more changes. Here's what it looks like now:

Screenshot 2025-04-17 at 2 21 26 PM

And if you're already using premium:

Screenshot 2025-04-17 at 2 24 01 PM

src/renderer/src/views/shared_components/UpsellComponent.vue
```diff
@@ -0,0 +1,253 @@
...
+                                    </template>
+                                    <template v-else>
+                                        Start a Team
+                                    </template>
+                                </p>
+                            </div>
+                        </div>
+                    </div>
+                    <div class="col-md-5 upsell-col">
+                        <div class="card text-center" @click="donateClicked">
```
SaptakS, line 170: I feel like these should be buttons instead of div to make them keyboard focusable. Right now they are not focusable from keyboard. But I am also not a huge fan of adding structured data inside a button. So maybe the "Click to donate" should be a button, so that at least something is focusable through keyboard?

SaptakS (changes requested):
I have just gone through the code and left some HTML comment. Might not be high priority but I think we should fix.

src/renderer/src/views/shared_components/UpsellComponent.vue
micahflee: I changed the whole `.card` divs to be buttons instead, and confirmed they can be focused with the keyboard. I agree it's not ideal to put structure data inside buttons (though I did notice that X and Facebook do it lol). But I think it's better this way, since the whole card is the clickable element, not just the text at the bottom.

SaptakS: X and Facebook are very bad in their UI so I won't consider them an inspiration. I would probably use `<span>` instead of `<div>`, since that is more corect. but I can do like a HTML validation and accessibility concerns pass on the entire tool and fix or catch these then.

SaptakS (approved):
Looks good to me.

PR #505, opened 2025-04-18
Fix up X's archiveBuild, and static site that is generated

### LINKED ISSUE #460: Investigate loading SQLite3 databases directly in static archive sites
Each account added to Cyd has its own sqlite database. When building the static website, it runs a bunch of queries to select the data that's needed and then saves it all together in one large JSON object, saved in a `.js` file that can be imported in the static website.

This means that we end up with two copies of the data: the sqlite database, and `assets/archive,js` that contains a copy of it in JSON format.

If we can, I think it would be simpler (and smaller file sizes for archives, and less code to maintain) if the static website could load data from the sqlite database directly. We could actually complete delete the `buildArchive` step if we did this.

The reason I didn't do this is because it's not clear to me how feasible it is. We'd need a sqlite library implemented in javascript and loaded into the browser, and we'd need to figure out how to load the sqlite db file from disk, in a local static html page.

But I just noticed that this site https://lite.datasette.io/ lets you specify the URL of a sqlite database, and it appears to fetch it and then start querying it.

Is it all in the browser? And can we fetch a sqlite file from the local filesystem? I don't know. If we can, I think this would be a solid improvement.

micahflee:
This will close #460.

I spent a lot of time looking into making SQLite in the browser work, and even implemented it, before concluding that it's actually worse.

But some of the changes I made were good changes still though, so rather than cherry-picking the good stuff, I just undid all of my changes again, and then squashed it all into a single commit. So, sorry about the one huge commit.

Here's what's changed in the static website:

- I removed a lot of repeated code. Tweets, Retweets, Likes, and Bookmarks are all different ways of displaying tweets. So I made all of these views use a shared TweetDisplayComponent instead.
- When viewing Messages, when it lists your conversations, it now hides your own user as a participant (because of course you participated in all conversations).

Here's what I changed in the Cyd code:

- I made a new `src/account_x/archive.ts` file that contains all of the logic related to selecting data from the SQLite database and serializing it. I also greatly cleaned up this logic and made it use far fewer SQL queries. I moved the `writeJSONArray` and `writeJSONObject` helper functions into `util.ts`, because FB and other platforms will need them too.
- When XAccountController closes the database it forces a WAL checkpoint, which basically writes all unwritten changes to the main database file
- The `archiveBuild` IPC function is considerably smaller and more readable, basically calling a `saveArchive` function defined in the new `archive.ts`, and then unzipping the static site zip


archive-static-sites/x-archive/src/components/TweetsDisplayComponent.vue
```diff
@@ -0,0 +1,42 @@
...
+        const tweetUsername = tweet.username ? tweet.username.toLowerCase() : '';
+        return tweetText.includes(lowerCaseFilterText) || tweetUsername.includes(lowerCaseFilterText);
+    });
+});
+</script>
+
+<template>
+    <div class="tweets-container">
+        <div class="filter-container">
+            <p><input type="text" v-model="filterText" class="form-control" :placeholder="filterTextPlaceholder"></p>
```
SaptakS, line 31: Probably needs to be addressed in a separate accessibility audit of the entire app, but inputs should be associated with a label. Placeholders are not good labels.

SaptakS (approved):
looks good to me

PR #511, opened 2025-04-25
wip: segment users across newsletters

redshiftzero:
The corresponding server change is here: https://github.com/lockdown-systems/cyd-server/pull/139

This segments users across two newsletters, one for free and one for premium users.


redshiftzero:
I'm going to rework the server side which should obviate the need to make changes on the client, so closing this

PR #512, opened 2025-04-27
wip: migrate to Bluesky without X/Twitter login

### LINKED ISSUE #413: X: ability to migrate to Bluesky without logging into an X account
**Is your feature request related to a problem? Please describe.**

In order to migrate from X to Bluesky, you need to first add an X account. You can only do this if you current actually still have an X account. If you've already deleted your X account, but you have an X archive, there's no way to migrate to Bluesky since you can't login to X.

**Describe the solution you'd like**

Maybe a good solution is, when you add a new X account, when it's prompting you to login to X, there should be a button to have limited functionality and just have the ability to import an X archive and migrate your data.

**Additional context**

From Bluesky: https://bsky.app/profile/denbib.bsky.social/post/3liojfauzqs2i

Image

redshiftzero:
Closes #413

TODO:
- [ ] Figure out how to handle the sidebar, perhaps only show the sidebar once we have a local database/archive?
- [ ] Modify the import page because the "Back" button goes back to a page that has many other options that aren't relevant / require login
- [ ] Test that the import process works without account info

### DIFF
--- a/src/renderer/src/view_models/XViewModel.ts
+++ b/src/renderer/src/view_models/XViewModel.ts
@@ -2987,9 +2987,9 @@ After you build a local database of your tweets, I can help you migrate them int
                 case State.WizardArchiveOnlyDisplay:
                     this.showBrowser = false;
                     this.instructions = `
-**You've chosen to save only your tweets and bookmarks.**
+**You've chosen to use a pre-existing X archive.**
 
-**I'll help you save them to your local database.**`;
+I'll help you save them so you can view them locally or migrate them to Bluesky.`;
                     await this.loadURL("about:blank");
                     this.state = State.WizardArchiveOnlyDisplay;
                     break;
--- a/src/renderer/src/view_models/XViewModel.ts
+++ b/src/renderer/src/view_models/XViewModel.ts
@@ -63,6 +63,8 @@ export enum State {
     WizardMigrateToBluesky = "WizardMigrateToBluesky",
     WizardMigrateToBlueskyDisplay = "WizardMigrateToBlueskyDisplay",
 
+    WizardArchiveOnlyDisplay = "WizardArchiveOnlyDisplay",
+
     RunJobs = "RunJobs",
 
     FinishedRunningJobs = "FinishedRunningJobs",
@@ -2840,7 +2842,9 @@ Hang on while I scroll down to your earliest bookmarks.`;
                     this.actionString = `Hello, friend! My name is **Cyd**. I can help you save and delete your tweets, likes, and direct messages from X.`;
                     this.instructions = `${this.actionString}
 
-**To get started, log in to your X account below.**`;
+**To get started, you can either:**
+- Log in to your X account below to access all features
+- Or click "Import Archive Only" to import your existing X archive without logging in`;
                     this.showBrowser = true;
                     this.showAutomationNotice = false;
                     await this.login();
@@ -2978,6 +2982,16 @@ You'll be able to access it even after you delete it from X.
 
 After you build a local database of your tweets, I can help you migrate them into a Bluesky account.`;
                     this.state = State.WizardMigrateToBlueskyDisplay;
+                    break;
+
+                case State.WizardArchiveOnlyDisplay:
+                    this.showBrowser = false;
+                    this.instructions = `
+**You've chosen to save only your tweets and bookmarks.**
+
+**I'll help you save them to your local database.**`;
+                    await this.loadURL("about:blank");
+                    this.state = State.WizardArchiveOnlyDisplay;
                     break;
 
                 case State.FinishedRunningJobs:
--- a/src/renderer/src/views/x/XView.vue
+++ b/src/renderer/src/views/x/XView.vue
@@ -34,6 +34,7 @@ import XWizardCheckPremium from './XWizardCheckPremium.vue';
 import XWizardMigrateBluesky from './XWizardMigrateBluesky.vue';
 import XFinishedRunningJobsPage from './XFinishedRunningJobsPage.vue';
 import XWizardSidebar from './XWizardSidebar.vue';
+import XWizardArchiveOnly from './XWizardArchiveOnly.vue';
 
 import XDisplayTweet from './XDisplayTweet.vue';
 
@@ -274,6 +275,19 @@ emitter?.on(`x-reload-media-path-${props.account.id}`, async () => {
 });
 
 const startJobs = async () => {
+    if (model.value.account.xAccount == null) {
+        console.error('startJobs', 'Account is null');
+        return;
+    }
+
+    // If in archive-only mode, skip authentication checks
+    if (model.value.account.xAccount.archiveOnly) {
+        await model.value.defineJobs();
+        model.value.state = State.RunJobs;
+        await startStateLoop();
+        return;
+    }
+
     // Premium check
     if (model.value.account?.xAccount && await xRequiresPremium(model.value.account.id, model.value.account.xAccount)) {
         // In open mode, allow the user to continue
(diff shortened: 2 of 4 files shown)


micahflee (changes requested):
I have some ideas for how to make the login state look better so the speech bubble isn't so tall, but I'm not entirely sure, and I think I'll just review this for functionality first, and maybe make wording suggestions in a separate PR...

When I click "Import Archive Only", here's what it looks like:

Screenshot 2025-07-04 at 3 43 22 PM

The sidebar says "Cyd could not detect how many likes and tweets you have." It only hides the "Try again" button in archive only mode, but I think it should hide the whole <p> tag.

And in fact, the Local Database and Migrate to Bluesky buttons don't really make sense here either (and Local Database shouldn't be possible to get to from archive only mode anyway). They're going to go away when https://github.com/lockdown-systems/cyd/pull/538 is merged, so maybe it makes sense to fix that later. But I think before importing from archive is complete, maybe the entire sidebar should be hidden?

Another thing I noticed is that if you hit an error importing (like this one, I tried importing an older twitter export format I have which) there's no way to recover:

Screenshot 2025-07-04 at 4 09 57 PM

I also think #538 might help with this because it introduces breadcrumbs and buttons components, and the breadcrumbs are displayed on this page -- though they'll have to change for archive only mode.

When you successfully import your archive, it's also not clear what you're supposed to do next.

---

For changes, this is what I'd recommend:

- [ ] hide the whole "Cyd could not detect how many likes and tweets you have." `<p>` tag in the sidebar in archive only mode
- [ ] if you're in archive only mode and you have never imported (see `xGetLastImportArchive` in `util_x.ts`), hide the whole sidebar
- [x] merge the `536-dashboard` branch (for PR #536) into this one

After merging, some of these problems will go away, but there will still need to be other changes. Here's what the dashboard looks like:

Screenshot 2025-07-04 at 4 19 27 PM

- [ ] in the dashboard, hide "Local Database" and "Delete from X"
- [ ] in the dashboard, add a new "Import X Archive" card that goes directly to XWizardArchiveOnly -- this will also need picking out a new royalty free graphic to use from https://www.flaticon.com/search?word=import&color=color and giving credit at https://cyd.social/credits/
- [ ] make sure the breadcrumbs and buttons for every page are correct in archive only mode, because they might need tweaking

redshiftzero:
Thanks for the review! I've rebased this on latest after the UX overhaul in #538 and applied the suggestions here

micahflee:
@redshiftzero I just pushed a bunch of commits. Here's what I've added.

- If you had a logged-in X account, it displayed the "Import X Archive" on the dashboard. This is now only displayed in archive-only mode.
- I re-ordered it so "Import X Archive" comes before "Migrate to Bluesky", and I also made it so if you've never imported, it says "Start Here".
- I made the ArchiveOnly view pretty, and used the ButtonsComponent
- If you started an archive-only account, then instead of importing went to the dashboard, then went to migrate to Bluesky, it was possible to then get to Local Database, which should be off limits. I fixed this, and overall updated the XLastImportOrBuild component.
- On the X login page, there's a waitForURL loop. I made it gracefully quit that loop if you click the archive only button.

Please review my commits.

I also ran into a problem where when I tried importing an X archive for `@micahflee`, it would fail at the "importing tweets" step with an error message "No tweets file found" while looking in a `deleted_account_` path. My `Documents/Cyd Dev/` folder had several `deleted_account_` dirs, as well as a `micahflee` dir. I deleted the `micahflee` dir and all of the `deleted_account_` dirs, and then tried again, and this time it worked without an error. So I can't reproduce it anymore.

Do you think this is worth digging into more? Or do you think we're good?

redshiftzero:
Your changes look good, I tried to reproduce the issue you reported but was unable to hit that issue. 

In the course of that I realized one case we should handle (will add to this PR) which is:
1. User already has an existing account for handle `@foo`.
2. They start a new profile, and attempt to import an "archive only" account that is also for `@foo`. The initial data folder will be called `deleted_user_x` and then will be renamed to `foo` once the import completes successfully. 

This could lead to data destruction in a case where the existing account already has data, so we should error in this case and get the user to remove their old profile if they still want it.

micahflee:
I hit a bug while testing this:

- I opened Cyd
- I already had one X account added, but I tried adding another
- I clicked the archive only button, and it threw an error

Screenshot 2025-07-30 at 3 06 55 PM

Details:

```json
{
  "error": "URL changed from https://x.com/login to about:blank\n\nURLChangedError: URL changed from https://x.com/login to about:blank\n    at Proxy.loadURLWithRateLimit (http://localhost:5173/src/view_models/XViewModel.ts:348:19)\n    at async Proxy.login (http://localhost:5173/src/view_models/XViewModel.ts:448:5)\n    at async Proxy.run (http://localhost:5173/src/view_models/XViewModel.ts:2147:11)\n    at async startStateLoop (http://localhost:5173/src/views/x/XView.vue:124:9)\n    at async http://localhost:5173/src/views/x/XView.vue:328:11",
  "state": "WizardArchiveOnly",
  "jobs": [],
  "currentJobIndex": 0
}
```

Context:

```json
{
  "logs": [
    {
      "timestamp": "2025-07-30T22:05:37.090Z",
      "func": "run",
      "message": "running state: WizardArchiveOnly"
    },
    {
      "timestamp": "2025-07-30T22:05:37.090Z",
      "func": "loadBlank"
    },
    {
      "timestamp": "2025-07-30T22:05:37.090Z",
      "func": "waitForLoadingToFinish",
      "message": "waiting for loading to finish"
    },
    {
      "timestamp": "2025-07-30T22:05:39.485Z",
      "func": "loadURL",
      "message": "URL loaded successfully"
    },
    {
      "timestamp": "2025-07-30T22:05:39.485Z",
      "func": "waitForLoadingToFinish",
      "message": "waiting for loading to finish"
    },
    {
      "timestamp": "2025-07-30T22:05:39.515Z",
      "func": "waitForLoadingToFinish",
      "message": "loading finished"
    },
    {
      "timestamp": "2025-07-30T22:05:39.522Z",
      "func": "waitForLoadingToFinish",
      "message": "waiting for loading to finish"
    },
    {
      "timestamp": "2025-07-30T22:05:39.688Z",
      "func": "waitForLoadingToFinish",
      "message": "loading finished"
    },
    {
      "timestamp": "2025-07-30T22:05:39.688Z",
      "func": "loadURLWithRateLimit",
      "message": "URL loaded successfully"
    },
    {
      "timestamp": "2025-07-30T22:05:39.688Z",
      "func": "loadURLWithRateLimit",
      "message": "checking if URL changed"
    },
    {
      "timestamp": "2025-07-30T22:05:39.690Z",
      "func": "loadURLWithRateLimit",
      "message": "UNEXPECTED, URL change to about:blank"
    }
  ],
  "currentURL": "about:blank"
}
```

micahflee:
Okay a few new things:

I fixed [the bug above](https://github.com/lockdown-systems/cyd/pull/512#issuecomment-3137952555). I realized you trigger it if you click "Import Archive Only" right away, before the X login page finishes loading.

If you look at `XWizardImportingPage`, it calls IPC functions in this order:

- `unzipXArchive`, if it's zipped, and stores the `unzippedPath`
- `verifyXArchive` (and `deleteUnzippedXArchive`) if it fails
- `importXArchive` to import tweets (passing in `unzippedPath`)
- `importXArchive` to import likes (passing in `unzippedPath`)
- `archiveBuild`
- `deleteUnzippedXArchive`

The code that renames the temp username folder to the username folder was in `importXArchive`, but that IPC method is called multiple times, and `unzippedPath` changes between them. So instead, I moved the rename folder code into `verifyXArchive`, because that only gets called once, and then I update the `unzippedPath` before doing the import.

There was also various code that tries to detect if it's in archive only mode based on if the username is null. And initializing archive only mode was called multiple times, including in the initDB function. I ripped all that out and made it so we explicitly call a new `initArchiveOnlyMode` IPC function, ensuring that the username will definitely be saved by the time we need it.

I also did a few other things. [This commit](https://github.com/lockdown-systems/cyd/pull/512/commits/baa3d627ea1eadf9626e21281f15622557b95f2f) makes the typescript IPC stuff way more DRY, and [this commit](https://github.com/lockdown-systems/cyd/pull/512/commits/9b7583234a4132ce216c0ff0b2a74d4e041fa19e) improves logging, so that debug logs are only shown if CYD_DEV=1.

Here's the main test plan I was trying to solve, and that now seems to work:

- Delete `~/Library/Application Support/Cyd Dev/` and `~/Documents/Cyd Dev/X`
- Create a new folder in `~/Documents/Cyd Dev/X` for an account you have an X archive of
- Run Cyd, adding an account in archive only mode, and try importing the archive
- It should give an error saying the folder already exists and to delete it
- Click back, delete the folder, and then try importing it again
- It should work this time

redshiftzero:
I've tested the latest commits, all LGTM!

micahflee (approved):

PR #534, opened 2025-06-23
Upgrade npm deps

### LINKED ISSUE #531: Update all npm dependencies, carefully
I used `npm-check-updates` to update all packages, and it turns out a lot of things broke subtly but spectacularly, so I rolled them back.

We should still do this, but with more care.

micahflee:
Fixes #531, most of the way.

Our npm dependencies were getting outdated, and so I've carefully updated all except for the electron-forge ones. Those still at version 7.4.0. Here's the output of `ncu`:

```
$ ncu  
Checking /Users/user/code/lockdown-systems/cyd/package.json
[====================] 75/75 100%

 @electron-forge/cli                         7.4.0  →  7.8.1
 @electron-forge/maker-deb                   7.4.0  →  7.8.1
 @electron-forge/maker-dmg                   7.4.0  →  7.8.1
 @electron-forge/maker-rpm                   7.4.0  →  7.8.1
 @electron-forge/maker-squirrel              7.4.0  →  7.8.1
 @electron-forge/maker-zip                   7.4.0  →  7.8.1
 @electron-forge/plugin-auto-unpack-natives  7.4.0  →  7.8.1
 @electron-forge/plugin-fuses                7.4.0  →  7.8.1
 @electron-forge/plugin-vite                 7.4.0  →  7.8.1
 @electron-forge/publisher-s3                7.4.0  →  7.8.1
```

When I upgrade to 7.8.1, Cyd appears to work. However, when making a binary (`npm run make-dev`) and then running it, it gives an error about `mime-types` not importing, but this actually means that none of the dependencies are getting imported correctly. I have not fixed that issue, but everything else should work.

I updated the atproto dependencies:

```
"@atproto/api": "^0.15.16",
"@atproto/oauth-client-node": "^0.3.1",
```

This caused a crash in `XAccountController.blueskyClientFromClientID`, because we were using `NodeOAuthClient.fromClientId` and it got removed in version 0.3.0 of that dependency: https://github.com/bluesky-social/atproto/blob/main/packages/oauth/oauth-client-node/CHANGELOG.md?plain=1#L15 -- so I worked around it.

I also updated `@electron/notarize` and `@electron/osx-sign` to new major versions. Before in `forge.config.ts`, I have had some hacky code to deal with macOS code signing and notarization, using preMake and postMake hooks. I deleted that, because the new updates seem to have fixed it.

# Test plan

- delete `node_modules` and run `npm install`
- tests should pass (both `npm run test` and `npm run test:components`)
- you should be able to run Cyd and use features like saving tweets (`npm run start`)
- you should be able to create a build (`npm run make-dev`), and then run the binary
- using the binary, you should be able to connect to a Bluesky and migrate tweets/delete those migrated tweets

It's hard to test the code signing and notarization code, but as part of this PR I made a Cyd Dev 1.1.14-dev release, and here is the Mac binary: https://releases.lockdown.systems/cyd/dev/macos/universal/Cyd%20Dev%201.1.14-dev.dmg -- you should be able to download this and then confirm that the app bundle is both code signed and notarized, and it should run on a Mac (either Intel or ARM64) without any security warnings.

# Still left to do

- Upgrade the electron-forge packages to the latest without the binaries it creates crashing
- Upgrade the npm deps all the way in `archive-static-sites`

I think we can leave these for future PRs.


micahflee:
> `npm run start` doesn't work for me in fedora directly anymore, because of a [seemingly known reported issue](https://github.com/electron/electron/issues/46538). I can still run, using the workaround `npm run start -- -- --gtk-version=3`. I am not sure if that is a particular concern to not update to electron 36

Okay, I have downgraded to electron 35.5.1, just to be safe.

I also updated the container that builds the linux releases to use node 22 instead of 20. If you want to check if the linux releases will actually work from the package manager, instead of the source tree, you should build a package manually the same way the Github Actions workflow does:

```sh
docker build -t my-builder --load .
docker run --rm -v $(pwd):/workspace -w /workspace my-builder ./scripts/make-linux-release.sh
```

This will create a .deb and .rpm in `out/make/deb/x64/*.deb` and `out/make/rpm/x64/*.rpm`.

SaptakS (approved):
Looks good to me

PR #538, opened 2025-06-30
Add X dashboard, and overhaul UI

### LINKED ISSUE #536: UX: make a dashboard page instead of menu sidebar
When I started making the UI, I designed a wizard for guiding the user through the steps of saving data and then deleting it, but it's gotten bigger and more complicated. Now there are three main pages:

- Local Database (build from scratch, import, archive)
- Delete from X
- Migrate to Bluesky

There will be more options too (Tombstone, Migrate to Mastodon, etc.).

When you open Cyd it starts in Local Database. When you're finished with Local Database, you may want to delete your data, but you also might want to migrate it, or neither.

I added on a nav menu in the sidebar to switch between these. But these sidebar items don't make sense for when #413 is implemented.

I think a better UX would be to remove the nav from the sidebar and instead create a dashboard page, with links to the relevant pages. When you're done with a task, it brings you back to the dashboard.

micahflee:
Fixes #536.

# What's different

I've changed the look and feel of the speech bubble. Now, the min-height of the speech bubble is the same as the height of the Cyd avatar (instead of being much smaller), with text vertically aligned. And instead of highlighting the call-to-action text by making it bold, I'm highlighting it by using an h1. Here's what the login page looks like, when adding a new X account:

Screenshot 2025-06-29 at 3 46 14 PM

Then you get to the Dashboard:

Screenshot 2025-06-29 at 3 52 51 PM

If the local database doesn't have any data yet, it has a "Start Here" badge in the corder of Local Database. If you do have data, that badge doesn't appear.

It used to be that when a Cyd account loads, it goes to the last page in the wizard you were most recently using (the local database page, delete page, or migrate page). I changed it so now it always just starts at the dashboard.

**Should we keep it like this? Or should the user still start from where they last left off?**

Also, the dashboard page uses the same royalty free icons we're using on the https://cyd.social/ homepage from Flaticon, with the license being free for personal or commercial use with attribution. If you go to the about page, there's now a link to Credits: https://cyd.social/credits/ -- we can slightly reword this here to apply to the app as well as the website.

I've updated the UI for all of the wizard pages. There are several changes:

- The right sidebar is much narrower, and I've removed the navigation from it
- The back button is always in the top left, and secondary outline
- The next button(s) are always in the bottom right, and primary
- The content in the middle, between the back button and the next button, now has its own dedicated scrollbar (I'll show in screenshots below)

Here's the local database page:

Screenshot 2025-06-29 at 3 58 13 PM

If you choose "Build data from scratch" and click "Continue to Build Options", here's the build options page:

Screenshot 2025-06-29 at 3 58 48 PM

Here's the review page for this action:

Screenshot 2025-06-29 at 3 59 39 PM

Here's what it looks like while saving tweets -- the speech bubble has a big H1 in it now, has a bigger min-height, and is centered vertically.

Screenshot 2025-06-29 at 3 59 59 PM

Here's the finished page:

Screenshot 2025-06-29 at 4 01 24 PM

Note that the next button is always in the very bottom right, even if the content doesn't take up that much space. Also, the narrow right sidebar has its own scrollbar to display a single column of stats now.

Here's the delete options page:

Screenshot 2025-06-29 at 4 07 47 PM

Notice that the back and next buttons are still always there, while the scroll bar section is just the form in the middle. I think this is a lot better -- before, you couldn't see the continue button at all without scrolling down, and I think this could have caused some confusion.

Here's what the review page looks like:

Screenshot 2025-06-29 at 4 08 28 PM

And finally, the migrate to Bluesky page is similar. It only shows the next button after you connect to a Bluesky account:

Screenshot 2025-06-29 at 4 08 59 PM

Also, I made one small unrelated packaging change. When run `npm run make-dev` (or any of the other `make-` commands) in macOS it not longer code signs and notarizes. It only does that when publishing a release. This fixes a regression related to updating all of the dependencies, and lets me make dev builds quicker (which I had to do to test Bluesky integration).

# Test plan

- Add a new X account and then test every single page of the wizard.
- Each page should look good with the new changes (back button at top left, next buttons at bottom right, content in the middle scrolling). 
- Make sure I didn't break anything in the UI.
- Give me recommendations on design or text changes.

# What's left to do

- This feature requires updating the screenshots in the documentation. I'll wait until it has a first review before adding that, just to avoid doing extra work if I end up making more changes.
- Update the text at https://cyd.social/credits/


src/renderer/src/views/x/XWizardDashboard.vue
```diff
@@ -0,0 +1,119 @@
...
+<template>
+    <div class="wizard-content">
+        <div class="wizard-scroll-content">
+            <div class="dashboard row g-3">
+                <div class="col-12 col-md-6 col-lg-6">
+                    <div class="card position-relative" @click="emit('setState', State.WizardDatabase)">
+                        <span v-if="!hasSomeData" class="start-here-badge badge bg-primary">Start Here</span>
+                        <div class="card-body align-items-center">
+                            <img src="/assets/icon-database.png" alt="Local Database">
+                            <h1>Local Database</h1>
```
SaptakS, line 38: Nitpick: I would make these `h2`. Ideally a screen should have a single h1.

micahflee:
Okay check it out.

I made the dashboard have 3 columns instead of 2. If you make the window really narrow is goes down to 2 still. I also made the heights all the same.

Screenshot 2025-07-02 at 3 43 34 PM

I also changed the `h1`s in the cards to `h2`s. I also made the main card text not have `text-muted`, and added last ran times in another `p` that is `text-muted` and centered (just for database and delete). It just says "Last ran" instead of getting into more granular detail -- like it doesn't say if you imported from an archive or built from scratch, etc.

I changed the style of the back buttons from `btn-outline-secondary` to just `btn-secondary` so they're solid grey now. I think it looks better...

Screenshot 2025-07-02 at 3 45 26 PM

micahflee:
Check it out. I've made two new components, `BreadcrumbsComponent` and `ButtonsComponent`, and I use them on all of the wizard pages. I also made a new util function `getBreadcrumbIcon` to make it so we can easily change icons across the board in one place.

Here's what it looks like:

Screenshot 2025-07-04 at 3 21 22 PM

The `BreadcrumbsComponent` takes a list of buttons, as well as a label and icon for the current page. In this screenshot, the list is just Dashboard.

The `ButtonsComponent` takes a list of back buttons and a list of next buttons, and it defines the back and forward icons  itself. Also, for the next buttons, you can make them disabled or hidden.

Here's another screenshot with more breadcrumbs:

Screenshot 2025-07-04 at 3 23 03 PM

The breadcrumbs list this time is Dashboard, Local Database, and Build Options, with the current page being Review. The back button in the buttons component at the bottom just goes back one page, while the breadcrumbs can bring you all the way back to any step you want.

Here's the Migrate to Bluesky options page:

Screenshot 2025-07-04 at 3 24 19 PM

In the buttons component here, if you have zero tweets to migrate, the "Continue to Review" button is disabled. There's also a "Delete Migrated Tweets from Bluesky" button, but you can't see it because it's hidden when the number of migrated tweets is zero.

Screenshot 2025-07-04 at 3 25 49 PM

In the screenshot above, all tweets that can be migrated have already been migrated, so there are zero tweets to migrate and the "Continue to Review" button is disabled. However there are > 0 tweets that have already been migrated, so the "Delete Migrated Tweets from Bluesky" button is not hidden.

Want to give final feedback, and if it all looks good, I will update the docs and then we can merge?

micahflee:
Okay finally and completely ready! I made two small changes:

- increased the bottom margin under the breadcrumbs
- made ButtonsComponent optionally support danger-styled buttons, so the button to delete posts from Bluesky is red instead of blue

And I've totally redone the docs. I remade all of the screenshots, and I rewrote parts of the text where it made sense, including adding a Dashboard page. I also updated the docs npm deps.

To test the docs:

```sh
cd docs
npm install
npm run start
```

Load http://localhost:3000/ and change from version 1.1.x to Next, and you can see all the docs. (When making the release, I need to remember to copy the docs from next to 1.1.x).

PR #544, opened 2025-07-10
Switch from npm to pnpm, and redo the linting and the GUI tests

### LINKED ISSUE #541: Consider switching from npm to pnpm
The `npm run` scripts make heavy use of deleting `node_modules` and reinstalling them, and pnpm is much more quick and efficient (and takes less disk space) than npm. I think if we switch to pnpm it will be a nicer development experience.

micahflee:
Fixes #541.

I started by switching from npm to pnpm. This project is actually a monorepo though with 4 different node projects:

- the root is Cyd
- `archive-static-sites/x-archive` is the X archive site
- `archive-static-sites/facebook-archive` is the Facebook archive site
- `docs` is the docs site

These different packages are defined in `pnpm-workspace.yaml`. With pnpm, there's only one `node_modules` and `pnpm-lock.yaml` file for all of the different workspaces. You can run commands in other workspaces like this, to build the `x-archive` site:

```sh
pnpm --filter x-archive build
```

I also deleted all of the linting packages from the archive sites, because the root workspace can handle linting for the whole git repo. This led me to spend a bunch of time fixing up the linting. I've improved the `eslint`. I found and fixed several new linting errors.

I also added `prettier` and reformatted _all of the files_, which is why so many files are modified.

For GUI tests, I ripped out cypress and replaced it with straight vitest, which makes the testing situation much simpler. You can now run all tests with just `pnpm test`, and the GUI ones (the few there were) have been rewritten.

## To test it out

Make sure you install `pnpm` with (`npm install -g pnpm`). Then install modules for the first time:

```sh
rm -rf node_modules
pnpm install
```

Run tests, lint, start Cyd, etc:

```sh
pnpm test
pnpm lint
pnpm fix # fix eslint errors
pnpm format # format with prettier
pnpm config-open
pnpm start
```

Create a binary, and make sure the binary runs:

```sh
pnpm run make-dev
```

Do these tests on Windows, Mac, and Linux.

## Notes

In Windows, when you run `pnpm run make-dev`, it creates a binary that successfully runs, but the command ends in an error. I believe this error is related to publishing a release though, and it won't happen when we actually run `pnpm run publish-dev`, so I feel reasonably comfortable with dealing with this during the next release...

Also in Windows, the tests don't pass yet. I opened a new separate issue for this: https://github.com/lockdown-systems/cyd/issues/558


SaptakS (approved):
Looks fine to me.

PR #560, opened 2025-08-14
Fix migrating media in Bluesky

micahflee:
This fixes a few issues:

- There were some typescript warnings related to atproto types which are fixed.
- There was an error using the `image-size` module to get aspect ratios when migrating images. I've removed that dependency (yay one less dependency) and replaced it with a simple function that looks like dimensions for PNGs and JPGs.
- There was an error migrating videos. I'm not sure when this changed, but for some reason, mp4 video mime types were being reported as "application/mp4" but atproto requires the mime type to be "video/mp4". So I just made it use "video/mp4" if the mime type was "application/mp4", and that fixes it.
- When migrating a tweet with multiple animated gifs in it, this now only migrates the first animated gif. It's because 1) animated gifs in X are technically mp4 videos, not gifs, and 2) atproto only supports one video per post.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,6 @@
     "glob": "^11.0.3",
     "http-encoding": "^2.1.1",
     "http-mitm-proxy": "^1.1.0",
-    "image-size": "^2.0.2",
     "jsdom": "^26.1.0",
     "marked": "^16.1.2",
     "mhtml2html": "^3.0.0",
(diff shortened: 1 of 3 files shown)


redshiftzero (approved):
LGTM:
- [x] In a test X account, make several posts that includes photos, videos, and animated gifs, including one post with multiple animated gifs.
- [x] Try migrating them to Bluesky. It should all work without errors.

Noting that the multiple animated gif post become a single animated gif on Bluesky as expected

Issue #582, opened 2025-09-08
Bulk delete all Facebook posts


redshiftzero:
**Is your feature request related to a problem? Please describe.**
Part of #380
In @SaptakS's recent #545 (rebased in #561) PR, we attempted to delete individual wall posts.

**Describe the solution you'd like**
A simpler approach is we could try to bulk delete posts by:
1. Navigating to the profile
2. Clicking "Manage Posts"
3. Repeatedly scrolling down and clicking "Select all" until 50 posts are selected
4. Click "Next"
5. Click "Delete Posts"
6. Click "Done"
7. Return to step 2 until all posts are deleted

redshiftzero:
Closed by #611

PR #611, opened 2025-12-01
Bulk delete Facebook posts

micahflee:
Implements #582, #617, #618.

This adds basic support for the Facebook platform. Some of the bigger things to note:

- Deleting everything from your Facebook wall is a premium feature. The full premium flow should be tested, with the dev server.
- Facebook support is hidden behind a flag, for now
- This relies on English-language strings, and so therefore when you delete wall posts, it first updates your language to English (if it isn't already), deletes posts, then changes it back
- This implements `BaseViewModel.clickElementByXPath` which makes development much faster, because in chrome/firefox dev tools, when inspecing an element, you can copy it's XPath
- This implements `BaseViewModel.safeExecuteJavaScript` which I think is a better way to execute js in the webview going forward, since it wraps it in all the error catching we need
- This submits Facebook progress records to the Cyd server via the API, and relies on a new API endpoint, see https://github.com/lockdown-systems/cyd-server/pull/150

# Feature flag

You need to enable the Facebook feature flag:

```sh
export CYD_FEATURE_FACEBOOK=1
npm run start
```

# Archive instructions

In the Facebook dashboard, clicking "Get Archive from Meta" links to https://docs.cyd.social/docs/facebook/get-archive, which is a broken link. To. view the instructions:

```sh
cd docs
npm run start
```

Then load http://localhost:3000/docs/next/facebook/get-archive.

# Testing

Here are some of the things to test for, though you might think of others:

- [ ] UX design and the language used should all look good
- [ ] Thoroughly test the premium gating. This entails running Cyd in local (`npm run config-local`), and running a separate instance of cyd-server from [this PR branch](https://github.com/lockdown-systems/cyd-server/pull/150).
    - [ ]  Facebook platform should be gated behind a premium account
    - [ ] The sign in/upgrade flow should all work fine -- if not signed in, it should prompt you to. If signed in but not paying for premium, it should give you instructions, etc.
    - [ ] If using open mode (`npm run config-open`) it should allow you to bypass premium gating
- [ ] Thoroughly test the deleting wall posts functionality
    - [ ] It deletes 50 at a time. So it should work fine to delete many more than this, e.g. 110 should take 3 rounds of deletions
    - [ ] It should work when there's stuff that cannot be deleted in the list of items (like avatar updates)
    - [ ] It should count the number of posts deleted successfully, and it should submit the correct number to the server
- [ ] Ensure that the database migrations work as expected (I actually think this might be an issue, but haven't tested a lot myself -- start a Cyd database using the latest release tag, then switch to this branch and make sure there are no errors)

# Future work before we're ready

This is a huge PR so I'm going to stop for now, but there are a few things that I think we also need to handle, which I haven't really tackled here:

- nevermind I did this: ~Handle rate limits. I don't have enough data in my test Facebook account to hit a rate limit, so I actually have no idea how Facebook handles them. Here's the issue: https://github.com/lockdown-systems/cyd/issues/617~
- nevermind I did this: ~Handle error reports. If something breaks (like it's trying to click a button that doesn't exist) it silently fails, but really it should submit an error report. Here's the issue: https://github.com/lockdown-systems/cyd/issues/618~
- Do we want more features in this first release? For example, I noticed that in the Manage Posts feature you can bulk hide things that aren't posts (updating your avatar or cover image, for example). We could have an option to automatically hide those. Also, you can bulk untag yourself from posts, though I'm not sure how this works and would require more testing. We should decide if we want to add these (because they might be easy wins?) or release with just deleting posts.

### DIFF
--- a/src/renderer/src/i18n/locales/en.json
+++ b/src/renderer/src/i18n/locales/en.json
@@ -343,6 +343,7 @@
   },
   "platform": {
     "u2fNotice": "If you use a U2F security key (like a Yubikey) for 2FA, press it when you see a white screen.",
+    "u2fNoticeFacebook": "If you use a U2F security key (like a Yubikey) for 2FA, press it when prompted.",
     "readMore": "Read more",
     "importArchiveOnly": "Import Archive Only (for deleted accounts with an archive)"
   },
--- a/src/renderer/src/views/PlatformView.vue
+++ b/src/renderer/src/views/PlatformView.vue
@@ -168,7 +168,11 @@ const currentJobsLength = computed(() => props.currentJobs.length);
         class="u2f-info text-center text-muted small ms-2"
       >
         <i class="fa-solid fa-circle-info me-2" />
-        {{ t("platform.u2fNotice") }}
+        {{
+          config.name === "Facebook"
+            ? t("platform.u2fNoticeFacebook")
+            : t("platform.u2fNotice")
+        }}
         <a
           v-if="config.urls.u2fDocs"
           href="#"


src/renderer/src/config/platforms/FacebookPlatformConfig.ts
```diff
@@ -1,49 +1,36 @@
...
-    hasU2FSupport: false, // Facebook does not support U2F security keys in our implementation
+    hasArchiveOnly: false,
+    hasPremiumGating: true,
+    hasComplexImport: false,
+    hasMigration: false,
+    hasU2FSupport: true,
   },
-
   urls: {
     helpDocs: "https://docs.cyd.social/docs/facebook",
```
redshiftzero: Nit: In the `PlatformConfig`, if we don't also populate `u2fDocs` in the URLs, we get a weird artifact:

Screenshot 2026-01-16 at 11 09 53 AM

So we can either populate that URL, or just modify this logic in `src/renderer/src/views/PlatformView.vue` to elide the final period if we don't have an informative URL:

```
       <!-- U2F security key notice (conditional on feature flag) -->
      <p
        v-if="
          config.features.hasU2FSupport && modelState == PlatformStates.Login
        "
        class="u2f-info text-center text-muted small ms-2"
      >
        <i class="fa-solid fa-circle-info me-2" />
        {{ t("platform.u2fNotice") }}
        <a
          v-if="config.urls.u2fDocs"
          href="#"
          @click="openURL(config.urls.u2fDocs)"
          >{{ t("platform.readMore") }}</a
        >.
      </p>
```

micahflee:
Thanks for the initial reviews, @redshiftzero @SaptakS! I've made a few changes.

I removed the CYD_FEATURE_FACEBOOK flag code, so that in this branch the Facebook features are always available. It seems that once we're done reviewing and merging, the next step is to make a dev release for Facebook support, and we shouldn't require the feature flag for that.

I fixed the database migrations. Please confirm by doing this:

- Delete `~/Library/Application Support/Cyd Dev/` (or the equivalent settings path on your platform)
- Switch to the `main` branch, open Cyd, and sign into an X account for good measure, then quit Cyd
- Switch to this branch and open Cyd again
- The migration should work without error

@redshiftzero, in regards to https://github.com/lockdown-systems/cyd/pull/611#discussion_r2699145357, I fixed the issue so it doesn't show an extra period. But can you also test what happens when your Facebook account uses U2F for 2FA, to determine whether or not we need to create instructions? (I just tried, but I'm having trouble setting my test FB account to use a Yubikey with a FIDO2 PIN set -- I could try again later maybe when I have access to another Yubikey.)

@SaptakS, regarding https://github.com/lockdown-systems/cyd/pull/611#pullrequestreview-3673668466, I think I fixed this:

- I made it so Cyd only deletes up to 10 posts at a time
- After each batch of posts, it waits 3 seconds (for good measure) and then refreshes the page (forcing it to reload new posts)

To test this, manually make many posts, such as 50, and then run the delete job. Make sure the number of posts deleted matches the number of posts you created.

I also fixed a problem I discovered on really slow internet connections. After clicking "Manage posts", it now waits up to 30 seconds for the posts to actually appear. I wouldn't worry about manually testing this.

@redshiftzero, I couldn't reproduce the language issue here https://github.com/lockdown-systems/cyd/pull/611#pullrequestreview-3671664788. Want to just try again and see if it's intermittent on your end?

In summary, here are no things to test:

- [ ] Make sure the facebook feature flag isn't required
- [ ] The migrations should work now
- [ ] Confirm that when u2fDocs isn't set, there isn't an extra period, and determine how Facebook U2F support works to figure if we need to add docs about it
- [ ] Confirm that it deletes all posts in a single run
- [ ] Ensure that the language issue was just a fluke
- [ ] Finish everything else in the review

redshiftzero:
Confirming the migrations issue is fixed. I am still seeing the language selection issue that means I can't proceed further in the deletion process, will try to debug that a bit since it's on my side. 

One other thing I noticed: if you create a Facebook account in Cyd and then leave the login page before entering your credentials, the session eventually times out. This is probably the case (?) on X so can be filed as a followup perhaps.

redshiftzero (approved):
I've tested the U2F flow, also tested via going through a couple of VPN endpoints that users can solve the CAPTCHA they get served and still login. Tested multi-page delete, that the count of posts were accurate, and avatar updates were skipped. 

LGTM!

PR #650, opened 2026-04-15
Fix fb deletion priority

### LINKED ISSUE #649: [Facebook] fix the deletion algorithm
The way the deletion algorithm is implemented is document in https://github.com/lockdown-systems/cyd/issues/643#issuecomment-4185306488, but there's an issue with the algorithm itself. Posts might get hidden when they should get deleted or untagged instead.

It prioritizes `delete`, then `untag`, then `hide`.

Let's say we have the posts with the following possible actions listed in reverse chronological order under Manage Posts:

1. hide
2. hide, delete
3. hide, delete

The existing algorithm works like this:

- load profile page and click manage posts
  - select 1, and see the highest priority action is `hide`
  - select 2, and see the highest priority action is still `hide` (it won't show `delete`, since 1 can't be deleted)
  - select 3, and see the highest priority action is still `hide` (it won't show `delete`, since 1 can't be deleted)
  - click next, choose `hide`, choose done
  - _1, 2, and 3 are hidden_

This is a bug. What we want it 1 to be hidden, and 2 and 3 to be deleted.

I think we maybe need to start by checking and unchecking each box, and recording the possible actions. Then, once we have mapped out the actions, bulk checking and doing the action.

SaptakS:
Fixes #649 

This PR does things slightly differently from the approach listed in the issue.

- It loops over the batch actions in priority (delete > untag > hide)
- It tries to select every item in the current priority batch action (delete in beginning and then so on) instead of determining the priority batch action from the first item
- If there is 0 items that the current priority action can be applied to, then it moves to the next action in the priority list, else it breaks from the loop and performs the action

This does make things slightly slower than before, but I think classifying (as suggested in the issue) would make it equally slow. I have added some test to replicate the issue and also tried to replicate the issue in a test account and in both cases, it seems to work. We may have to brainstorm later if we can improve performance, but this definitely seems to be working in deleting, untagging and hiding everything.

I think this process is also easier, if we want to provide a better summary of how many elements were deleted, how many were untagged and how many were hidden.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,8 +10,7 @@
       "hasInstallScript": true,
       "license": "proprietary",
       "workspaces": [
-        "archive-static-sites/x-archive",
-        "docs"
+        "archive-static-sites/x-archive"
       ],
       "dependencies": {
         "@atproto/api": "^0.18.21",
--- a/src/account_facebook/controller/stats/getProgressInfo.ts
+++ b/src/account_facebook/controller/stats/getProgressInfo.ts
@@ -19,8 +19,25 @@ export async function getProgressInfo(
     totalWallPostsDeleted = parseInt(totalWallPostsDeletedConfig);
   }
 
+  const totalWallPostsUntaggedConfig: string | null =
+    await controller.getConfig("totalWallPostsUntagged");
+  let totalWallPostsUntagged: number = 0;
+  if (totalWallPostsUntaggedConfig) {
+    totalWallPostsUntagged = parseInt(totalWallPostsUntaggedConfig);
+  }
+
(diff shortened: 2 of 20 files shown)


micahflee (changes requested):
This works great! I think we should add two additional things to this PR:

- Separate the progress into items deleted, untagged, and hidden, instead of just deleted. (This will also require a small `cyd-server` change, when submitting progress.)
- Display more verbose progress info, possibly in the speech bubble? Instead of just saying "I'm removing all posts from your Facebook wall." it should give granular updates:
  - When it starts looping through a different action, it should say something like, "I'm looking for a batch of posts to **delete/untag/hide**..."
  - After it finds a batch, it should say something like, "I'm **deleting/untagging/hiding** X posts..."

PR #658, opened 2026-06-05
Fix X archive-only UX

micahflee:
The archive-only mode button was confusing.

Now, when you add a new account X account, it pops up a new modal asking if you can still access your X account:

Screenshot 2026-06-05 at 11 22 51 AM

If you choose "I Can Login to X", it loads the login page (which no longer has the archive only button).

If you choose "I Can't Login, Use Archive Only Mode" it goes to the archive only mode page.


redshiftzero (approved):
nice! good improvement, the UX was a bit janky there

PR #657, opened 2026-06-05
Allow for more URL changes during login

micahflee:
I got a report from a Cyd user that 1) they were trying to login to their X account and it wasn't detecting it and 2) the UI for archive-only mode was confusing.

This PR addresses the first one. I'm not reproducing the issue, but I think this fixes it anyway, based on the error report they submitted. The login function includes:

```
  // Load the login page and wait for it to redirect to home
  await vm.loadURLWithRateLimit("https://x.com/login", [
    "https://x.com/home",
    "https://x.com/i/flow/login",
  ]);
```

It loads `https://x.com/login`, and the other two URLs (`https://x.com/home` and `https://x.com/i/flow/login`) are expected to load. Once it finally makes it to `https://x.com/home`, it concludes that the login was successful.

Based on the error report, it actually tried loading `https://x.com/i/jf/onboarding/web?mode=login`, which triggered an error. It seems that X changed their login flow, at least for this user. So to keep things simpler and more robust, I've changed it so that any URL that starts with `https://x.com/i/` is an expected URL and won't throw an error. This should allow URLs like `https://x.com/i/jf/onboarding/web?mode=login` through, and allow for completing the login.

This is from the error report that they submitted:

```json
{
  "logs": [
    {
      "timestamp": "2026-06-04T21:08:34.551Z",
      "func": "run",
      "message": "running state: Login"
    },
    {
      "timestamp": "2026-06-04T21:08:34.551Z",
      "func": "login",
      "message": "logging in"
    },
    {
      "timestamp": "2026-06-04T21:08:34.551Z",
      "func": "loadURLWithRateLimit",
      "message": [
        "https://x.com/login",
        [
          "https://x.com/home",
          "https://x.com/i/flow/login"
        ],
        false
      ]
    },
    {
      "timestamp": "2026-06-04T21:08:34.552Z",
      "func": "waitForLoadingToFinish",
      "message": "waiting for loading to finish"
    },
    {
      "timestamp": "2026-06-04T21:08:34.605Z",
      "func": "domReadyHandler",
      "message": "dom-ready"
    },
    {
      "timestamp": "2026-06-04T21:08:34.753Z",
      "func": "waitForLoadingToFinish",
      "message": "loading finished"
    },
    {
      "timestamp": "2026-06-04T21:08:34.753Z",
      "func": "loadURL",
      "message": "try #0, https://x.com/login"
    },
    {
      "timestamp": "2026-06-04T21:08:35.959Z",
      "func": "loadURL",
[...]

PR #659, opened 2026-06-05
Updates for 1.1.22

micahflee:
In preparation for the next release, this PR does the following:

- Updates all of the dependencies (this is most of the PR)
- Puts adding a Facebook account back behind a feature flag
- Adds a promo for Cyd for Mobile and Bluesky

With all of these changes (and particularly the dependency updates), everything should still work. To test:

- Make a build with `npm run make-dev` and run the final binary, rather than just testing in dev mode
- Do a smoke test: build an archive, test the archive works, delete something

Screenshot 2026-06-05 at 1 24 50 PM


redshiftzero:
I've been testing this a bit from the dev binary and I'm getting a hang before building my database (FYI I cherry picked both #658 and #657 prior to testing:

Screenshot 2026-06-06 at 8 09 24 PM

micahflee:
> I've been testing this a bit from the dev binary and I'm getting a hang before building my database (FYI I cherry picked both #658 and #657 prior to testing:
> 
> Screenshot 2026-06-06 at 8 09 24 PM

Hmm, I'm not able to reproduce. I created a dev build in both Mac and Linux, and it works fine for me.

On Mac I had recently deleted `~/Library/Application Support/Cyd Dev/` so I started with clean state. But in Linux I had an old state with already-signed in accounts, and this worked fine for me too.

I'm curious if there are console errors. Can you check, and open a new issue if you see them? I think though in order to get this release out sooner, I will merge this and make the next release.

PR #691, opened 2026-08-06
Fixes building the database from scratch in twitter

### LINKED ISSUE #669: Build database from scratch is not working for a few days now

**Describe the bug**
it does the regular routine of loading the twitter pages. but ends up with zero data being logged. this was fine till a few days ago. 

**To Reproduce**
Try to run the Build database from scratch wizard in app.
 
**Expected behavior**
It should scroll through the profile page and record tweets and retweets. In the past, it uses to sometimes end up with no data logged esp just after a cleanup. But running it after a delay or at a different time of the day usually worked. At this point, it just does not work. It even shows the profile page for a second or two. I have been checking it multiple times a day for the last couple of days. 

**Screenshots**

- OS: [e.g. Windows, Mac, or Linux] Windows
- Version of Cyd: [e.g. 1.0.25] 1.1.22

**Additional context**

SaptakS:
Fixes #669 

It actually almost fixes the issue. The two main reasons seemed to be that the DOM wasn't fully loading, and the response data didn't have some objects. 

However, the DM/conversations is completely broken. I did update the paths that they use now, but that itself doesn't fix the issue. I am still looking into it and will create a separate draft PR, with more information related to that.

### DIFF
--- a/src/account_x/controller/index/indexParseTweetsResponseData.ts
+++ b/src/account_x/controller/index/indexParseTweetsResponseData.ts
@@ -149,8 +149,6 @@ export function indexParseTweetsResponseData(
             entries.content.itemContent.tweet_results.result.core.user_results
               .result &&
             entries.content.itemContent.tweet_results.result.core.user_results
-              .result.legacy &&
-            entries.content.itemContent.tweet_results.result.core.user_results
               .result.core &&
             entries.content.itemContent.tweet_results.result.legacy
           ) {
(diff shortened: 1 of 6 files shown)


jonatino:
@SaptakS do you think this also fixes cyd not archiving bookmarks properly anymore when using this option? 

It logs in, open bookmarks then instantly says it's finished without fetching anything new. This stopped working about 2 weeks ago.

image

SaptakS:
@jonatino I tested bookmarks also, and have adjusted the sleep time for that as well. Locally testing, this PR seems to fix the bookmarks also.

SaptakS:
Based on https://github.com/lockdown-systems/cyd/issues/693 I think there is another reason/step why building database from scratch might fail. I will check if I can add that fix/fallback also to this PR.

redshiftzero:
Trying this out, and for some reason with my test account I'm getting into an infinite loop when trying to save tweets:

Screenshot 2026-08-22 at 9 23 49 PM

redshiftzero:
Ah maybe I'm also hitting #693

SaptakS:
@redshiftzero Yeah. I think I need to look into #693. With the chat when they updated the route, they also updated the MITM paths that we track. So hopefully its not the same. I will try to take a look this week into it.

micahflee (approved):
not finished but merging so far

PR #692, opened 2026-08-06
Attempt to fix XChat (Twitter DMs)

SaptakS:
Seems like ever since twitter moved to XChat (or if they moved user to XChat), seems like our DM parsing is broken. I have updated the selectors in this PR, but seems like our MITM logics for particular JSON are not functional anymore, since those API are not called anymore. I have looked into it for some time and I haven't been able to figure out a replacement GraphQL query that we can MITM. So we might need to actually parse the page DOM to get the conversations URL if we want to do that


micahflee:
closing this, because I think supporting X Chat is much more work than just tweaking our existing DM code, and there's a good argument for just no longer supporting it anyway

